PR #5, opened 2022-09-16
Add step-by-step jupyter notebook and default run directory 

ashjbarnes:
A notebook has been added that linearly steps through the process of creating a new regional domain. It uses modified versions of Angus' scripts as well as code that performs the bash commands Angus used to set up EAC and PANAN. A default run directory has been included, which is copied by the notebook into the new run directory and modified to suit the newly created experiment.


angus-g (changes requested):
Thanks! As we discussed, I think this is one of the two possible approaches:

1. develop as a library, and import that into the notebook which describes how to use the library
2. use [nbdev](https://nbdev.fast.ai/) and make the notebook(s) the main point of development, exporting to a library for other consumption

Happy to take perspectives either way on this; I think having a library created through some means is important, and the documentation from the notebook is useful. If we decide to go for the first option, we can look at the organisation of functions and files, and wrapping everything up into an actual package.

### Minor points
For the PR itself, you should remove the `.ipynb_checkpoints`, `__pycache__` and `original_scripts` directories. I like the idea of having an example run that actually exercises the outputs from the regional domain (and tide) generation. Maybe that would be better situated in its own repository though?

.gitignore
```diff
@@ -0,0 +1,2 @@
+__pycache__
+.ipy*
```
angus-g, line 2: This would catch notebooks too?

ashjbarnes: thanks Angus, fixed!

ashjbarnes:
I've collected all of the functions into three files: The ones we've written, the ones that we've borrowed from the NWA repo, and the HCtFlood tool. I guess that we could turn our functions into a proper module, but given that some of it is still hardcoded for our purposes (particularly the fact that it's designed to work with ACCESS-OM2) I thought that for now it might make sense to leave it as a set of functions that could be modified as needed by the user

angus-g:
So I think we discussed a plan to move the notebook portion of this into https://github.com/COSIMA/cosima-recipes, then make this repository hold just a Python package (with the goal of being included in the COSIMA conda environment) to do the Python backend stuff.

PR #13, opened 2023-05-18
Inputagnostic

ashjbarnes:
Ok I left this branch for way too long without merging. This is the future branch of mom6 regional following a full refactor in which functions are made agnostic to model inputs and wrapped up in classes. The 'regional_library.py' is where current scripts live. 'mom6_regional_scripts.py' is legacy and no longer needed after regrid runoff is put into pipeline


ashjbarnes:
hold on I'll clean some things up first then re open...

ashjbarnes:
thx @aekiss

aekiss (approved):

PR #23, opened 2023-06-02
Delete `legacy_README.md` + updates `README.md`

navidcy:
No need to keep that there. Github saves it in the repositorie's history for eternity!


ashjbarnes (approved):
thanks Navid!

PR #34, opened 2023-06-14
Some cleanup

navidcy:
Just some organizing.

@ashjbarnes, what does `regional_model_scripts.py` includes? What's the difference from the `mom6-regiona.py`?


angus-g:
I'm not sure it makes sense to move that into the library. Probably better to just delete it, we can excavate it from history if ever needed.

navidcy:
Oh, you said the magic word!

ashjbarnes:
The only reason the legacy mom6_regional_scripts.py is still there is I've not incorporated the Regrid Runoff step. Perhaps we could delete it for now, and then alter the data table in the default rundirs so they're not expecting runoff at all? Make a note that runoff will / might be included at some point

navidcy:
@ashjbarnes, feel free to alter the data table in the default run directory in this PR?

navidcy:
@ashjbarnes shall I merge and you deal w runoff in a different PR?

ashjbarnes:
Oh good point - I think we should deal with it now because otherwise the demo notebooks won't work after merging.

How do I modify within the PR? Go back to my review and change some code?

navidcy:
locally checkout the branch and then commit to it!

```
git checkout ncc/move-regional-model-scripts
``` 

Makes sense?

ashjbarnes:
Ok I'm going to draw this one out a bit to make sure that on merging the demo notebooks work. I'll remove the runoff from the runs, modify notebook accordingly, and ensure that the demo still produces a runnable mom6 configuration

navidcy:
Yeah, do that!

navidcy:
I moved the black badge to a different PR; see #36

navidcy:
@ashjbarnes I'll let you merge this PR as soon as you have tested that the demo notebook run or whenever you are happy.

ashjbarnes:
Thanks Navid! I'm happy with this now. The demo notebooks could do with some more (or better thought out) markdown but they work fine for now. I'll save this for a future PR

PR #35, opened 2023-06-14
Fix `NaN` interpolation to catch edge cases and remove time dimension

ashjbarnes:
This fixes a few headaches in the initial condition interpolater. These are important for running with a newer executable where more netcdf variable type checks are done

### DIFF
--- a/mom6_regional/mom6_regional.py
+++ b/mom6_regional/mom6_regional.py
@@ -477,72 +477,9 @@ class experiment:
                 ),
                 sep="\n",
             )
-            return xr.open_dataset(self.mom_input_dir + "/hgrid.nc")
-
             return hgrid
 
-    def _old_make_hgrid(self):
-        """
-        Generates a mom6 hgrid ready to go. Makes a basic grid first, then uses FRE tools to create a full hgrid with all the metadata.
-        """
-        ## Total number of qpoints in x
-        qpoints_x = int((self.xextent[1] - self.xextent[0]) / self.res) + 1
-        qpoints_y = int((self.yextent[1] - self.yextent[0]) / self.res) + 1
-        if qpoints_x == 0 or qpoints_y == 0:
-            raise ValueError("Please ensure domain extents match chosen resolution")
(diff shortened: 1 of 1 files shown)


ashjbarnes:
Ok I ran black and then pushed again but the test fails and idk why

navidcy:
Seems that the first commit you did black was not even able to run!
https://github.com/COSIMA/mom6-regional/actions/runs/5264807245/jobs/9516578361#step:3:23

There were 2 parentheses in line 635 of this commit? [254d848](https://github.com/COSIMA/mom6-regional/pull/35/commits/254d8486d7b4ba2e10bf20a319411534cf7d1d5d)

navidcy:
That was a parse error -- I think not even python could parse that.

navidcy:
OK! I think you have some WEIRD characters at the end of your file!

navidcy:
I think it originated because you had a parsing error with 2 parentheses in your first commit....

mom6_regional/mom6_regional.py
```diff
@@ -1503,3 +1457,4 @@ def brushcut(self, ryf=False):
             )
 
         return segment_out, encoding_dict
+  
```
navidcy, line 1460:
```suggestion

```

navidcy: There were some WEIRD characters here. If I opened with VS Code look what I saw:

Screenshot 2023-06-15 at 10 56 46 am

ashjbarnes: Yeah I saw that and thought I fixed it with the second commit. That's not present in my local copy (black actually ran fine for me) but maybe I forgot to save and commit again. Just re-committed now

navidcy:
That's good...
But I'd have to trust you that you fixed a bug. This is the type of things that a test should have caught ;)

navidcy (approved):

navidcy:
Gotcha!

navidcy:
Note: If this was part of dealing with #34 then just feel free to commit there. :)

PR #37, opened 2023-06-15
Remove placeholder `parabola` func/test

navidcy:
This removes the placeholder test for `parabola` function and adds a meaningful test.

### DIFF
--- a/tests/test_trivial.py
+++ b/tests/test_trivial.py
@@ -1,13 +1,6 @@
-from mom6_regional import parabola
+import numpy as np
+from mom6_regional import angle_between
 
-
-def func(x):
-    return x + 1
-
-
-def test_answer():
-    assert func(4) == 5
-
-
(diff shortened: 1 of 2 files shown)


tests/test_trivial.py
```diff
@@ -1,13 +1,7 @@
...
-
-def test_answer():
-    assert func(4) == 5
-
-
-def test_parabola():
-    assert parabola(2) == 4
+# placeholder trivial test test
+def test_angle_between():
+    assert angle_between([1, 0, 0], [0, 1, 0], [0, 0, 1]) == np.pi / 2
```
angus-g, line 7: Maybe use `np.isclose`? Not usually reliable to exactly compare floats

navidcy: thanks!

angus-g (approved):
Cool, all good now!

PR #38, opened 2023-06-15
Build API docs

angus-g:
The page for the moment needs a decent bit of work for prettiness, but it's doing automatic API documentation. I will add the demo notebook as its own built page. This uses MyST so it should allow for standalone Markdown pages, rather than rST. Docs should be building on RTD from now too.


navidcy (approved):
This looks great!

Can we add a docs badge at the `README.md` that points to the documentation?

navidcy:
merge?

PR #49, opened 2023-06-22
Update instructions for local docs built

### LINKED ISSUE #48: Local docs don't see the package itself?
So following the directions at https://regional-mom6.readthedocs.io/en/latest/contributing.html#documentation to build the docs locally the API website is empty.

Seems like the docs don't see the package itself? If I do `pip install ../.` then it works; but every time I edit a docstring I need to re-install? Somehow I'd expect (intuitively) that `sphinx` would be able to monitor changes in the package itself? Am I doing something wrong or that's how it is?

(Because the docs are not built via a Github Action I don't know exactly what's the pipeline for building them...)

cc @angus-g

navidcy:
Closes #48

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -47,10 +47,16 @@ cd docs
 conda create --name docs-env --file requirements.txt
 ```
 
-Then activate this environment and run `make`:
+Then activate this environment and install the package itself as an editable install (`pip install -e`).
 
 ```{code-block} bash
 conda activate docs-env
+pip install -e ../.
+```
+
+Now we can make the docs
+
+```{code-block} bash
 make html
 ```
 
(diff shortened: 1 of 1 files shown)


docs/contributing.md
```diff
@@ -47,10 +47,16 @@ cd docs
 conda create --name docs-env --file requirements.txt
 ```
 
-Then activate this environment and run `make`:
+Then activate this environment and install the package itself as an editable install (`pip install -e`).
 
 ```{code-block} bash
 conda activate docs-env
+pip install -e ../.
```
angus-g, line 54: This could just be `pip install -e ..`. Also, I guess the install instruction should be updated for the non-conda method too?

angus-g (approved):
Looks alright now. I assume you've tested the conda instructions.

navidcy:
Yeap, I'd done that ;)

Issue #39, opened 2023-06-15
`angle_between` and `quad_area` functions


navidcy:
I'm wondering if we really need to write our own functions for angle between vectors and the surface of a spherical quadrilateral. I have the gut feeling that there should be a python package that does that. Then we don't have to maintain and test these functions:

https://github.com/COSIMA/mom6-regional/blob/145e9f2b75b1009c70c4d00a4b32d14714870bf6/mom6_regional/mom6_regional.py#L264-L298

If we really need to rewrite those then a docstring is needed. ;)

navidcy:
https://github.com/COSIMA/mom6-regional/blob/145e9f2b75b1009c70c4d00a4b32d14714870bf6/mom6_regional/mom6_regional.py#L276 

seems redundant

navidcy:
I'd suggest

```python
v1xv2 = np.cross(v1, v2)
v1xv3 = np.cross(v1, v3)
cosangle = np.dot(v1xv2, v1xv3) / np.sqrt( np.dot(v1xv2, v1xv1) * np.dot(v1xv3, v1xv3)
angle = np.arccos(cosangle)
```

ashjbarnes:
Yeah good points Navid! That would be ideal. I just copied and pasted for now to be speedy and not get bogged down in details but this is certainly more elegant.

navidcy:
OK. Well we can first add tests for function as they are currently implemented and then changing will be trivial (if we break anything tests will fail).

navidcy:
Btw, as is defined at the moment, `angle_between` fails if  any of the vectors is the origin `[0, 0, 0]` (everything is zero and we end up dividing with 0).

ashjbarnes:
Golden rule of (Ashley’s) regional modelling: no poles

navidcy:
hm... (0, 0, 0) is the Earth's center, right?

navidcy:
It seems that the `angle_between` is only used by `quad_area`, right?

And that the `quad_are` is a function that computes the horizontal area of a cell centered at `(lat, lon)`. The shape of that cell is a spherical quadrilateral.

Perhaps we can utilize exact formulas for the area of spherical triangles. I recently coded these for the spherical grid in Oceananigans:
https://github.com/CliMA/Oceananigans.jl/blob/eb38eeade577eca5056b99ca8839ca6c674ae0e2/src/Grids/grid_utils.jl#L548-L664

[...]

PR #51, opened 2023-06-22
Some code cleanup

navidcy:
This was mostly done in an attempt to understand what the functions are doing.

There was somewhere a definition of 1 degree in radians and then used to multiply all lat, lon values. I presume that's simply to convert degrees to radians. I changed to use `np.deg2rad` which is self-explanatory. 

~Also I relaxed the restriction that longitudes needed to be equidistant and that the `dx = x[1] - x[0]`. Now `dx = np.diff(x)`; will this create issues?~

I haven't tested anything. @ashjbarnes have a look and please test? Supposedly the code should not be different so perhaps we can compare a grid constructed on `main` and the same grid on this PR to ensure that nothing broke?


regional_mom6/regional_mom6.py
```diff
@@ -295,65 +295,66 @@ def angle_between(v1, v2, v3):
...
+        Here it is assumed that `λ` and `φ` are on a perfectly rectangular grid.
+        Rotated grid needs to be handled separately.
+
+        Make sure both `λ` and `φ` *must* be monotonically increasing.
 
     Args:
-        x (numpy.array): All longitude points on supergrid. Assumes even spacing.
-        y (numpy.array): All latitude points on supergrid.
+        `λ` (numpy.array): All longitude points on the supergrid.
+        `φ` (numpy.array): All latitude points on the supergrid.
```
angus-g, line 337: Don't add Markdown to arguments docstrings like this, they are already recognised as variables from the context.

regional_mom6/regional_mom6.py
```diff
@@ -295,65 +295,66 @@ def angle_between(v1, v2, v3):
...
     ddd = (px * qx + py * qy + pz * qz) / np.sqrt(ddd)
     angle = np.arccos(ddd)
+
     return angle
 
 
 # Borrowed from grid tools (GFDL)
 def quad_area(lat, lon):
-    """Returns area of spherical quad (bounded by great arcs)."""
+    """Returns area of spherical quadrilaterals (bounded by great arcs)."""
```
angus-g, line 304: Guess this should get a proper docstring?

regional_mom6/regional_mom6.py
navidcy: OK -- didn't know!

Also what is "the super grid"? And if λ, φ are on the super grid then why we divide by 2 further down?

regional_mom6/regional_mom6.py
navidcy: Yes! I'm trying to understand first what it does.. there is some discussion in #39. :)

regional_mom6/regional_mom6.py
angus-g: The supergrid is the set of all cell centres *and* corners, so it kind of looks like a double-resolution tracer grid. I expect they're divided by 2 to get the normal resolution grid.

navidcy: oh I see!
so is the `λ, φ` inputs of this function the "supergrid" but then the output is the normal grid? then `dλ = np.diff(λ) * 2` (and not ` / 2`), right?

@ashjbarnes can elaborate perhaps?

angus-g: Hmm yeah, division does sound strange in that case.

regional_mom6/regional_mom6.py
```diff
@@ -295,65 +295,64 @@ def angle_between(v1, v2, v3):
...
+    R = 6371e3  # mean radius of the Earth
 
-    R = 6371000  # This is the exact radius of the Earth if anyone asks
-
-    # dx = pi R cos(phi) / 180. Note dividing resolution by two because we're on the supergrid
+    # dx = R * cos(φ) * np.deg2rad(dλ) / 2. Note, we divide dy by 2 because we're on the supergrid
     dx = np.broadcast_to(
-        np.pi * R * np.cos(np.pi * y / 180) * 0.5 * res / 180,
-        (x.shape[0] - 1, y.shape[0]),
+        R * np.cos(np.deg2rad(y)) * np.deg2rad(np.diff(λ)) / 2,
```
angus-g, line 347: This gained a `np.diff`? Does it change anything?

navidcy: was part of my questions to @ashjbarnes; I didn't understand why we weren't doing it previously.
the array should be the same but now the method also allows for non-uniform spacing in longitude... probably not necessary... I would definitely like to see a horizontal grid comparison using `main` and this PR before merging.

angus-g: Ah yeah, just making sure we have a comment at the right code location :)

navidcy: before this was just `np.diff(λ) -> res` which was equal to `x[1]-x[0]`

regional_mom6/regional_mom6.py
ashjbarnes: Sorry for such a late reply! I've totally abandoned this while I've been away. Was this resolved in the end? This code was all copied from the NWA people in a rush with the intention of rewriting later.

navidcy: There is no reason to apologize!

But, what is NWA?

ashjbarnes: NWA is the North West Atlantic experiment that the NOAA people have done. This is where some of the code has been borrowed from, esp for the grid gen https://github.com/jsimkins2/nwa25

regional_mom6/regional_mom6.py
ashjbarnes: Ok so on first pass this code fails when running the demo notebooks. Firstly there was a typo (y written instead of phi in line 347) but on fixing it I then get this error:

image

To test I cloned the demo notebook, checked that it still worked using the main branch then switched to the PR branch.

ashjbarnes (approved):

PR #56, opened 2023-06-23
Add demo notebooks to docs

angus-g:
Uses [nbsphinx](https://nbsphinx.readthedocs.io/) like COSIMA Recipes, but a bit more stripped down. Hopefully it'll just work as expected!


navidcy (approved):
This looks great!

PR #65, opened 2023-09-05
Simplify `angle_between` + refactor `quadrilateral_area`

### LINKED ISSUE #39: `angle_between` and `quad_area` functions

navidcy:
Uses a more geometrically-inspired definition for the `angle_between`. Also refactors `quadrilateral_area`. Now `quadrilateral_area` computes the area of _one_ quadrilateral and `quadrilateral_areas` computes the areas of all quadrilaterals on a lat-lon grid.

Closes #39


regional_mom6/regional_mom6.py
```diff
@@ -279,24 +278,16 @@ def dz(npoints, ratio, target_depth, min_dz=0.0001, tolerance=1):
...
-    return angle
+    """Returns angle v2-v1-v3 that is between the vectors v1-v2 and v1-v3."""
+
+    v1xv2 = np.cross(v1, v2)
+    v1xv3 = np.cross(v1, v3)
+    cosangle = np.dot(v1xv2, v1xv3) / np.sqrt(
+        np.dot(v1xv2, v1xv2) * np.dot(v1xv3, v1xv3)
+    )
+
+    return np.arccos(cosangle)
```
angus-g, line 290: Yeah, much clearer!

tests/test_grid_generation.py
```diff
@@ -6,6 +6,14 @@
...
 
 
+def random_unit_vector():
+    """Return a random unit vector on the unit sphere."""
+    z = 2 * np.random.rand(1)[0] - 1
+    z
+    θ = 2 * np.pi * np.random.rand(1)[0]
+    return [np.sqrt(1 - z**2) * np.cos(θ), np.sqrt(1 - z**2) * np.sin(θ), z]
+
+
```
angus-g, line 16: Did you mean to use this in a test?

angus-g (commented):
Changes to `angle_between` look good, but not sure if you intended to use the random vector generation in the test or not?

tests/test_grid_generation.py
navidcy: oh I used it to produce a bunch of random triplets of unit vectors to ensure that the new angle function gave me the same answer as the old one... but I should remove it now!

navidcy:
somehow the tests are failing tho... I'll figure it out

angus-g:
Yeah, it looks like that function is actually being called with some kind of array of vectors, so it needs to be multidimension-aware?

navidcy:
OK, I see the issue... 

It's that `quadilateral_area` gives `angle_between` 3D arrays... (e.g. arrays of `np.shape = (3, nx, ny)`).

How can I make it broadcast over the 2nd and 3rd dimension? I don't remember how to do it in python... :(

angus-g:
I guess we could do `np.cross(..., axis=0)` and `np.tensordot(..., axis=([0],[0]))`, but that's a bit ugly... It might also work if we use arrays of shape `(nx, ny, 3)` instead? That would also be the "correct" indexing convention for numpy -- having the vector axis first is the Fortran convention.

navidcy:
This

```python
    nv, nx, ny = np.shape(c0)

    c0 = np.reshape(c0, (nx, ny, nv))
    c1 = np.reshape(c1, (nx, ny, nv))
    c2 = np.reshape(c2, (nx, ny, nv))
    c3 = np.reshape(c3, (nx, ny, nv))
```

reshapes it... But then how do I ask the `angle_between` to broadcast over last two dimensions?

angus-g:
I think if you write it in such a way that the dot products and cross products are computed over the last axis, it'll just work as expected. I think that's the case for the cross product, from https://numpy.org/doc/stable/reference/generated/numpy.cross.html#numpy.cross:

> The cross product of a and b in is a vector perpendicular to both a and b. If a and b are arrays of vectors, the vectors are defined by the last axis of a and b by default, and these axes can have dimensions 2 or 3.

But I'm not sure what to do for the dot product...

navidcy:
So at the moment `quadrilateral_area` computes an array of areas.

I suggest we change it so that 

```python
def quadilateral_area(v1, v2, v3, v4):
    """Returns area of a spherical quadrilateral on the unit sphere that
    has vertices on 3-vectors `v1`, `v2`, `v3`, `v4` (counter-clockwise
    orientation is implied). The area is computed as the excess of the
    sum of the spherical angles of the quadrilateral from 2π."""

    a1 = angle_between(v1, v2, v4)
    a2 = angle_between(v2, v3, v1)
    a3 = angle_between(v3, v4, v2)
    a4 = angle_between(v4, v1, v3)

    return a1 + a2 + a3 + a4 - 2.0 * np.pi
```

that gives the area of a single quadrilateral. Then we have

```python
def quadilateral_areas(lat, lon):
    ....
```

that uses `quadilateral_area` to computes the quadrilateral areas for the grid.

How do you like that?

navidcy:
OK, I did it **but** it's not using broadcasting but it's doing the dreaded for-loop. See 

https://github.com/COSIMA/regional-mom6/blob/30254bd94229c5690954312e64fa51ee7ab61edb/regional_mom6/regional_mom6.py#L322-L329

@angus-g or @ashjbarnes can you convert it to broadcasting or something more efficient?

angus-g:
Yeah, I'll take a look.

navidcy:
@angus-g I added a validation in `quadilateral_area` via [d091f32](https://github.com/COSIMA/regional-mom6/pull/65/commits/d091f3271f88b6d3f23c5374a2d6afc4c1be8329) that checks if all vectors provided are of the same length. Can we add a test to ensure that en exception is raised when vectors of different length are provided?

navidcy:
> @angus-g I added a validation in `quadilateral_area` via [d091f32](https://github.com/COSIMA/regional-mom6/pull/65/commits/d091f3271f88b6d3f23c5374a2d6afc4c1be8329) that checks if all vectors provided are of the same length. Can we add a test to ensure that en exception is raised when vectors of different length are provided?

chatGPT told me how to do it! done via  aa5ed7a

ashjbarnes (approved):
I like it - very verbose and readable

regional_mom6/regional_mom6.py
```diff
@@ -279,47 +280,85 @@ def dz(npoints, ratio, target_depth, min_dz=0.0001, tolerance=1):
...
+    orientation is implied). The area is computed via the excess of the
+    sum of the spherical angles of the quadrilateral from 2π."""
+
+    if not (
+        np.isclose(np.dot(v1, v1), np.dot(v2, v2))
+        & np.isclose(np.dot(v1, v1), np.dot(v2, v2))
+        & np.isclose(np.dot(v1, v1), np.dot(v3, v3))
+        & np.isclose(np.dot(v1, v1), np.dot(v4, v4))
+    ):
+        raise Exception("vectors provided must have the same length")
```
angus-g, line 308: Better to be more specific and use `ValueError` than a general exception, I think.

navidcy: Done

navidcy:
> I like it - very verbose and readable

OK, just the optimization of the for-loop with broadcasting and we are ready to merge then :)

angus-g:
Here's an attempt at doing the broadcasting: https://github.com/COSIMA/regional-mom6/commit/2a5bce495230195652dbf538139604300d11c353

It's a little ugly to have to `dstack` everything in the tests, but on a 50x50 grid it takes 1.6ms/it vs. 715ms/it (400x speedup) vs. the for-loop.

navidcy:
Sounds great! Why don't you push to this branch?

PR #66, opened 2023-09-05
Simplify quadilateral_area + better doc

navidcy:
I changed to human numbering 1, 2, 3,...

I think the formula `a1 + a2 + a3 + a4 is exact` so no need to update that. I did update the docs a bit to be truthful to what `quadrilateral_area` does. Have a look @ashjbarnes, thanks!


navidcy:
Together with #65 this PR resolves #39.

ashjbarnes:
Looks good thanks Navid! Yeah this function is one I borrowed and was really opaque earlier

ashjbarnes (approved):

PR #70, opened 2023-09-06
Made the use of FRE tools optional, and stopped them being called on …

ashjbarnes:
Simple PR. I've just removed the calling of FRE tools (mosaic , check mask , ocean mask) from all other functions and put them in their own stand alone wrapper called FRE_tools. This way we can test all of the Python code before we've incorporated the FRE tools into the pipeline, and also write separate tests for the FRE tool compatibility.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -832,8 +832,7 @@ class experiment:
         fill_channels=False,
         minimum_layers=3,
         maketopog=True,
-        positivedown=False,
-        run_fretools = True
+        positivedown=False
     ):
         """Cuts out and interpolates chosen bathymetry, then fills
         inland lakes.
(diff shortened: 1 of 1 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -550,18 +550,6 @@ def _make_hgrid(self, gridtype):
...
-                " "
-            )
-            print(
-                "FRE TOOLS: Make solo mosaic\n\n",
-                subprocess.run(
-                    [self.toolpath + "make_solo_mosaic/make_solo_mosaic"] + args,
-                    cwd=self.mom_input_dir,
-                ),
-                sep="\n",
-            )
```
navidcy, line 564: so these lines are now called by some other function? where?

navidcy:
Can we have a better PR title? Seems like unfinished sentence at the moment?

navidcy (approved):

PR #73, opened 2023-09-13
Replace ESMF subprocess calls with new parallel implementation of xesmf (0.8.x) directly in Python

### LINKED ISSUE #72: Installation doesn't (appear to) correctly set up XESMF
I'm building tests for the regridding side of the module and have run into an issue: the xesmf regridder terminates with

`ESMF_FileRegrid Library needed by ESMF not present - Internal subroutine call returned Error`

This is discussed [here](https://github.com/JiaweiZhuang/xESMF/issues/47)

To reproduce: Install as per instructions on readme and then type

`pytest -v --pyargs xesmf`

and the 'regridding' parts of xesmf will fail

I'll finish writing the tests on my local copy where the installation has been fixed manually, then have a go at fixing the installation but since this was done by @angus-g I might need a bit of help

ashjbarnes:
This is a big PR that grew as a few threads where pulled. Namely, there are several problems with the main branch:

1. It relies on running the ESMF regridder as a subprocess with mpirun, since it's way faster than doing xe.Regridder(). However this is extremely janky, and it turns out that it doesn't work at all with the default installation instructions in the README. This means that the code simply doesn't work for people trying to use it who aren't on NCI and are able to rely on analysis3 (somehow their implementation of xESMF works with mpirun don't ask me why)

2. In writing my test function, I found some general bugs in my code which made some unhelpful assumptions. Where you see things being renamed in the `ocean_forcing` method , this is why. Basically the code worked by accident because the dims happened to fall in the right order, and I've changed it to ensure that it relies on the coordinate names rather than the axis.

3.  There was still some legacy stuff in the demo notebooks

This PR fixes these three things, as well as adding tests for the bathymetry and ocean forcing, updating the demo notebooks and readme to reflect changes. It goes together with #70 and means that a new user could download and install the package in a new environment and run the demo notebooks. 

This PR resolves #72


ashjbarnes:
Ok so looks like there are still some installation issues that cause the tests to fail. Namely the numpy installation isn't right, leading to the[ "module not found: bottleneck" issue](https://github.com/pydata/bottleneck/issues/380)

environment-ci.yml
```diff
@@ -2,4 +2,5 @@ name: mom6-regional-dev
 channels:
   - conda-forge
 dependencies:
-  - esmpy
+  - xesmf 
+  - bottleneck
```
navidcy, line 6: why is this bottleneck here?

regional_mom6/regional_mom6.py
```diff
@@ -596,7 +596,7 @@ def ocean_forcing(
         ## Do initial condition
 
         ## pull out the initial velocity on MOM5's Bgrid
-
+        print(subprocess.run("pwd", shell=True))
```
navidcy, line 599: remove?

navidcy, line 599:
```suggestion
```

regional_mom6/regional_mom6.py
```diff
@@ -797,7 +797,16 @@ def ocean_forcing(
...
         print("Saving outputs... ", end="")
-        vel_out.fillna(0).drop("time").to_netcdf(
+
+        ## Remove time IF it exists. Users may already have done so for us
+        if "time" in vel_out.dims:
+            vel_out = vel_out.isel(time=0).drop("time")
+        if "time" in vel_out.dims:
+            tracers_out = tracers_out.isel(time=0).drop("time")
+        if "time" in vel_out.dims:
+            eta_out = eta_out.isel(time=0).drop("time")
```
navidcy, line 807:
```suggestion
        for out in [vel_out, tracers_out, eta_out]:
            if "time" in out.dims:
                out = out.isel(time=0).drop("time")
```

regional_mom6/regional_mom6.py
```diff
@@ -886,39 +896,54 @@ def bathymetry(
...
-                bathy_path, chunks={varnames["xh"]: 500, varnames["yh"]: 500}
-            )[varnames["elevation"]]
+            if chunks != "auto":
+                chunks = {varnames["xh"]: chunks["lon"], varnames["yh"]: chunks["lat"]}
+
+            bathy = xr.open_dataset(bathy_path, chunks=chunks)[varnames["elevation"]]
 
             bathy = bathy.sel(
-                {varnames["yh"]: slice(self.yextent[0] - 0.1, self.yextent[1] + 0.1)}
+                {varnames["yh"]: slice(self.yextent[0] - 1, self.yextent[1] + 1)}
```
navidcy, line 911: Why do we need to remove either 0.1 or 1 from here?

navidcy: OK, I see.. you just take a bit more of the bathymetry than the domain.
I think we should either:
- have this buffer size be automatically adjusted based on the resolution the user provides + the resolution of the bathymetry
_or_
- at least have a comment here so that we know that there is a hard-coded assumption that 1 degree buffer is enough
-

regional_mom6/regional_mom6.py
```diff
@@ -886,39 +896,54 @@ def bathymetry(
...
             ).astype("float")
 
-            bathy = nicer_slicer(
-                bathy, np.array(self.xextent) + np.array([-0.1, 0.1]), varnames["xh"]
-            )
+            ## Here need to make a decision as to whether to slice 'normally' or with nicer slicer for 360 degree domain.
+
+            if bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][0] > 355:
```
navidcy, line 916:
```suggestion
            # if the grid goes around the globe then we want to use `nicer_slicer`
            if np.isclose(bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][0], 360):
```

regional_mom6/regional_mom6.py
```diff
@@ -886,39 +896,54 @@ def bathymetry(
...
+                ## Assume that we're dealing with a global grid, in which case we use nicer slicer
+                bathy = nicer_slicer(
+                    bathy,
+                    np.array(self.xextent) + np.array([-0.1, 0.1]),
```
navidcy, line 920: why 0.1?

regional_mom6/regional_mom6.py
```diff
@@ -886,39 +896,54 @@ def bathymetry(
...
-            bathy.elevation.attrs["coordinates"] = "lon lat"
-
-            bathy.to_netcdf(
+            bathyout = xr.Dataset({"elevation": bathy})
+            bathy.close()
+
+            bathyout = bathyout.rename({varnames["xh"]: "lon", varnames["yh"]: "lat"})
+            bathyout.lon.attrs["units"] = "degrees_east"
+            bathyout.lat.attrs["units"] = "degrees_north"
+            # bathyout.lon.attrs["_FillValue"] = 1e20
```
navidcy, line 938:
```suggestion
```

regional_mom6/regional_mom6.py
```diff
@@ -934,31 +959,84 @@ def bathymetry(
...
+
+            # rewrite chunks to use lat/lon now for use with xesmf
+            if chunks != "auto":
+                chunks = {"lon": chunks[varnames["xh"]], "lat": chunks[varnames["yh"]]}
 
+            tgrid = tgrid.chunk(chunks)
             tgrid.lon.attrs["units"] = "degrees_east"
             tgrid.lon.attrs["_FillValue"] = 1e20
             tgrid.lat.attrs["units"] = "degrees_north"
             # tgrid.to_netcdf(f"{self.mom_input_dir}tgrid.nc", mode = "w",engine='netcdf4')
```
navidcy, line 993: do we need this commented code?

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+            "lata": np.linspace(yextent[0] - 5, yextent[1] + 5, 100),
+            "lona": np.linspace(xextent[0] - 5, xextent[1] + 5, 100),
+        },
+    )
+    # name the bathymetry variable of xarray dataarray
+    bathy.name = "elevation"
+
+    bathy.to_netcdf("bathy.nc", mode="a")
+    bathy.close()
+    expt.bathymetry(
```
navidcy, line 82:
```suggestion

    # Now use this bathymetry as input in `expt.bathymetry()`
    expt.bathymetry(
```

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+    return
+
+
+import numpy as np
+import pytest
+from regional_mom6 import experiment
+import subprocess
+import xarray as xr
+
+# reload(experiment)
```
navidcy, line 101:
```suggestion
```

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+        vlayers,
+        dz_ratio,
+        depth,
+        mom_run_dir,
+        mom_input_dir,
+        toolpath,
+        gridtype,
+    )
+
+    ## Generate some bathymetry to test on
```
navidcy, line 161:
```suggestion
    ## Generate some initial condition to test on
```

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+
+    initial_cond = xr.Dataset(
```
navidcy, line 163:
```suggestion
    # initial condition includes, temp, salt, eta, u, v
    initial_cond = xr.Dataset(
```

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+                coords={
+                    "lata": np.linspace(yextent[0] - 5, yextent[1] + 5, 100),
+                    "lona": np.linspace(xextent[0] - 5, xextent[1] + 5, 100),
+                    "deepness": np.linspace(0, 1000, 10),
+                },
+            ),
+        }
+    )
+
+    eastern_boundary = xr.Dataset(
```
navidcy, line 212:
```suggestion

    # Generate boundary forcing

    eastern_boundary = xr.Dataset(
```

navidcy (approved):
ok! lgtm

did you ensure that the notebooks run now? I didn't check that!

there is some cleanup that can happen but it's not urgent for this PR!

README.md
```diff
@@ -61,3 +62,9 @@ pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de05
 ## Getting started
 
 The two example notebooks walk you through how to use the package on two different sets of input datasets. Make sure that you can get at least one of these working on your setup with your MOM6 executable, and then try to modify them to apply to your domain with your bathymethry / forcing / boundaries. 
+
+### Note for larger domains
+
+The example notebooks deal with a pretty small domain, but might not scale well if you try to run something really computationally expensive. When we first wrote this tool, xesmf (the package that handles all the regridding) didn't support parallelised weight generation. This meant that for large domains, you needed to call the underlying ESMF library with `mpirun`. This is especially important for the `.bathymetry` method, as the weight generation step becomes very slow for large areas if there's no spatial chunking. With the [introduction of `xesmf 0.8.x`](https://xesmf.readthedocs.io/en/latest/changes.html), a parallel option is now possible directly in python.
```
navidcy, line 68: if we require xesmf v0.8 or later, can't we add that compat requirement in the package dependencies? @angus-g?

navidcy: will it be something that has to do with this line?
https://github.com/COSIMA/regional-mom6/blob/d9b997f0e497d58db4a827432be7c88d2b472b1b/pyproject.toml#L16

README.md
```diff
@@ -18,6 +18,7 @@ If you find this package useful and have any suggestions please feel free to ope
 3. A bathymetry file that at least covers your domain
 4. 3D ocean forcing files *of any resolution* on your choice of A, B or C Arakawa grid
 5. Surface forcing files (eg ERA or JRA reanalysis)
+6. A copy of [GFDL's FRE tools](https://github.com/NOAA-GFDL/FRE-NCtools) downloaded and compiled somewhere.
```
navidcy, line 21:
```suggestion
6. [GFDL's FRE tools](https://github.com/NOAA-GFDL/FRE-NCtools) must be downloaded and compiled somewhere.
```

README.md
navidcy, line 68:
```suggestion
The example notebooks deal with a pretty small domains; they might not scale well if you try to run something really computationally expensive. When we started developing this package, xesmf (the package that handles all the regridding) didn't support parallelised weight generation. This meant that for large domains, users needed to call the underlying ESMF library with `mpirun`. This is especially important for the `.bathymetry` method, as the weight generation step becomes very slow for large domains if there's no spatial chunking. With the [introduction of `xesmf 0.8.x`](https://xesmf.readthedocs.io/en/latest/changes.html), a parallel option is now possible directly in python. 
```

environment-ci.yml
angus-g: This file shouldn't have been changed. bottleneck is needed for `interpolate_na` (from xarray) which was causing failing tests. But because the main library code uses it, we should bring it in as a dependency in `pyproject.toml`.

angus-g (changes requested):
Dependencies and CI environment need to be fixed, I haven't looked at the rest of the PR but I don't think this should be merged yet.

README.md
navidcy: Done

regional_mom6/regional_mom6.py
angus-g: This won't work, you will only be overwriting the `out` variable, not the "actual" variable. This is just re-implementing `.squeeze("time")`, probably just use that, i.e.

```python
vel_out.squeeze("time").fillna(0)...
```

etc.

regional_mom6/regional_mom6.py
```diff
@@ -934,31 +959,84 @@ def bathymetry(
...
+            # If we have a domain large enough for chunks, we'll run regridder with parallel=True
+            parallel = True
+            if len(tgrid.chunks) != 2:
+                parallel = False
+
+            try:
+                regridder = xe.Regridder(bathyout, tgrid, "bilinear", parallel=parallel)
+            except TypeError:
+                print(
+                    "Warning: xesmf parallel regridding failed. You likely have an older version of xesmf installed - try updating to 0.8.x or higher."
```
angus-g, line 1013: This is superfluous if we require `xesmf>=0.8` for the package.

regional_mom6/regional_mom6.py
```diff
@@ -934,31 +959,84 @@ def bathymetry(
...
+            #         shell=True,
+            #         cwd=self.mom_input_dir,
+            #     ).returncode
+            #     != 0
+            # ):
+            #     raise RuntimeError(
+            #         "Regridding of bathymetry failed! This is probably because mpirun was initialised earlier by xesmf doing some other regridding. Try restarting the kernel, then calling .bathymetry() before any other methods."
+            #     )
+
+            ## END CHANGES
```
angus-g, line 1036: Remove comments.
```suggestion

```

README.md
navidcy, line 68: I think a comment is needed here explaining what is "small" and what is "large" domain. (order of magnitude...?)

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+                    "lata": np.linspace(yextent[0] - 5, yextent[1] + 5, 100),
+                    "lona": np.linspace(xextent[1] - 0.5, xextent[1] + 0.5, 5),
+                    "deepness": np.linspace(0, 1000, 10),
+                    "time": np.linspace(0, 1000, 10),
+                },
+            ),
+        }
+    )
+
+    subprocess.run("mkdir dummyinputs", shell=True)
```
angus-g, line 266: Better to use a [`tmp_path`](https://docs.pytest.org/en/7.1.x/reference/reference.html#std-fixture-tmp_path) fixture. Resorting to `subprocess` should only be in dire circumstances.

README.md
navidcy: Also, do we need to give people the remark about the history of the development here? Can we, perhaps, avoid all the discussion since now our package requires xesmf v0.8 or later so that's sorted, right?

navidcy: Also, do we need to give people the remark about the history of the development here? Can we, perhaps, avoid all the discussion since now our package requires xesmf v0.8 or later so that's sorted, right?

tests/test_expt_class.py
```diff
@@ -0,0 +1,292 @@
...
+            "tracers": {"temp": "temp", "salt": "salt"},
+        },
+        boundaries=["east"],
+        gridtype="A",
+    )
+
+    print(subprocess.run("rm dummyinputs/ic_unprocessed", shell=True))
+    print(subprocess.run("rm dummyinputs/east_unprocessed", shell=True))
+    ## Make an IC file to test on
+
```
angus-g, line 291: With `tmp_path` (mentioned above), these should go.
```suggestion
```

README.md
angus-g: Agreed

regional_mom6/regional_mom6.py
navidcy: Nice catch

README.md
```diff
@@ -61,3 +62,9 @@ pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de05
...
+
+### Note for larger domains
+
+The example notebooks deal with a pretty small domain, but might not scale well if you try to run something really computationally expensive. When we first wrote this tool, xesmf (the package that handles all the regridding) didn't support parallelised weight generation. This meant that for large domains, you needed to call the underlying ESMF library with `mpirun`. This is especially important for the `.bathymetry` method, as the weight generation step becomes very slow for large areas if there's no spatial chunking. With the [introduction of `xesmf 0.8.x`](https://xesmf.readthedocs.io/en/latest/changes.html), a parallel option is now possible directly in python. 
+
+If you have the latest version of `xesmf` the package will attempt to regrid in parallel, and if not will throw a warning and run it in serial. It will also print out the relevant `mpirun` command which you could use as a backup. Depending on your setup, you may need to write scripts that implement the package to access more computational resources than might be available in a Jupyter notebook as is the case at the NCI facility.
```
navidcy, line 70: No need to say "if" since now by requirement users _will have to have_ xesmf v0.8 or later.

Also, remove anything related with NCI from here.

navidcy: NCI-related comments belong in the example at https://github.com/COSIMA/cosima-recipes/pull/216

navidcy (changes requested):
@ashjbarnes,

the docs seem to fail because the notebooks wanna use the curated conda environments that live on the NCI; see 

https://readthedocs.org/projects/regional-mom6/builds/21919690/

I presume you changed something in the notebooks?

navidcy:
I think the culprit for docs failing is here:

https://github.com/COSIMA/regional-mom6/blob/6f084ac5662292661dccfad93b78c88b35fa7e23/demos/model-forced.ipynb#L494

angus-g:
> I think the culprit for docs failing is here:
> 
> https://github.com/COSIMA/regional-mom6/blob/6f084ac5662292661dccfad93b78c88b35fa7e23/demos/model-forced.ipynb#L494

This line is the same in `main` though, so maybe not? Although the other notebook changed **from** that kernel back to `python3`...Maybe that has something to do with it? Haven't looked in detail...

navidcy:
Now we get the error `no kernel python3`... :)

angus-g:
At least it's consistent! :grin:

angus-g:
Maybe we (I) need to look at https://nbsphinx.readthedocs.io/en/latest/configuring-kernels.html

navidcy:
Are the notebooks actually evaluated or just displayed? I thought they were merely displayed in which case it wouldn't matter. But I'd be delighted if the cells are actually evaluated. This ensures that our docs are continuously up to date.

angus-g:
> Are the notebooks actually evaluated or just displayed?

That's a good question. I've set up [nbsphinx](https://nbsphinx.readthedocs.io/en/latest/index.html) in its default configuration -- it'll execute notebooks if they're committed without any output. But since the `model-forced.ipynb` had its outputs cleared, it's now trying to execute it to generate it fresh. I agree that'd be the way to do it! I guess we just need to look into what kernels are available in the RTD environment so it can actually run it properly...

regional_mom6/regional_mom6.py
navidcy: ok, then perhaps we can leave it as it was initially?
or just fix the typos (I believe there were few typos) 

```suggestion
        if "time" in vel_out.dims:
            vel_out = vel_out.isel(time=0).drop("time")
        if "time" in tracers_out.dims:
            tracers_out = tracers_out.isel(time=0).drop("time")
        if "time" in eta_out.dims:
            eta_out = eta_out.isel(time=0).drop("time")
```

navidcy: otherwise, if indeed the if statement should be the same for all, we can combine under one if:

```suggestion
        if "time" in vel_out.dims:
            vel_out = vel_out.isel(time=0).drop("time")
            tracers_out = tracers_out.isel(time=0).drop("time")
            eta_out = eta_out.isel(time=0).drop("time")
```

navidcy:
> > Are the notebooks actually evaluated or just displayed?
> 
> That's a good question. I've set up [nbsphinx](https://nbsphinx.readthedocs.io/en/latest/index.html) in its default configuration -- it'll execute notebooks if they're committed without any output. And since the `model-forced.ipynb` had its outputs cleared, it's now trying to execute it to generate it fresh. I agree that'd be the way to do it! I guess we just need to look into what kernels are available in the RTD environment so it can actually run it properly...

Great! Indeed, that would be the ideal!

regional_mom6/regional_mom6.py
angus-g: Probably the second suggestion -- not sure we can assume if one file has time, they all do? I will have a look at how to streamline all of this, not particularly clean in any configuration.

angus-g:
> @ashjbarnes does this PR resolve #72? If so, can you add the phrase "resolves #72" (without quotation marks) in the first comment of the PR so that the issue is automatically closed upon merging?

You can also link the issue (or others) from the *Development* section of the sidebar in a PR.

environment-ci.yml
ashjbarnes: Thanks Angus, yeah I was just trying different things to try and get it to work without throwing the bottleneck error

regional_mom6/regional_mom6.py
ashjbarnes: The reason for this is that it's not guaranteed that the user has already removed the 'time' dim from the initial conditions. EG, GLORYS dataset leaves time in, but when I made my dummy data there was no time dim so original implementation threw an error. I figured that if it's an "Initial condition" it would be intuitive that the user might delete the time dim so wanted to handle this

regional_mom6/regional_mom6.py
ashjbarnes: Yeah lets do the latter for now 
```suggestion
                {varnames["yh"]: slice(self.yextent[0] - 1, self.yextent[1] + 1)} #! Hardcoded 1 degree buffer around bathymetry selection. Should automatically select buffer in future
```

regional_mom6/regional_mom6.py
ashjbarnes: This will need testing on an example notebook. I'd rather save it for another small PR later if there's an easy way to flag that?

README.md
ashjbarnes:
```suggestion
6. [GFDL's FRE tools](https://github.com/NOAA-GFDL/FRE-NCtools) must be downloaded and compiled on the machine you are using.
```

README.md
ashjbarnes: Yeah agreed - we could remove all of this for now and revise the readme in a later PR

ashjbarnes:
> > Are the notebooks actually evaluated or just displayed?
> 
> That's a good question. I've set up [nbsphinx](https://nbsphinx.readthedocs.io/en/latest/index.html) in its default configuration -- it'll execute notebooks if they're committed without any output. And since the `model-forced.ipynb` had its outputs cleared, it's now trying to execute it to generate it fresh. I agree that'd be the way to do it! I guess we just need to look into what kernels are available in the RTD environment so it can actually run it properly...

Ah yeah that's my bad - maybe we should just re-run the notebooks then commit? I deleted the output since it's a bit messy but this could be cleaned up. e.g we don't need the dask dashboard and other clutter. Currently the only useful output in the notebooks is the plot of the topography anyway

navidcy:
> > > Are the notebooks actually evaluated or just displayed?
> > 
> > 
> > That's a good question. I've set up [nbsphinx](https://nbsphinx.readthedocs.io/en/latest/index.html) in its default configuration -- it'll execute notebooks if they're committed without any output. And since the `model-forced.ipynb` had its outputs cleared, it's now trying to execute it to generate it fresh. I agree that'd be the way to do it! I guess we just need to look into what kernels are available in the RTD environment so it can actually run it properly...
> 
> Ah yeah that's my bad - maybe we should just re-run the notebooks then commit? I deleted the output since it's a bit messy but this could be cleaned up. e.g we don't need the dask dashboard and other clutter. Currently the only useful output in the notebooks is the plot of the topography anyway

Nah.. Let's figure out how to actually have the notebooks run by the docs. That way the docs will be up to date by construction!

ashjbarnes:
> > > > Are the notebooks actually evaluated or just displayed?
> > > 
> > > 
> > > That's a good question. I've set up [nbsphinx](https://nbsphinx.readthedocs.io/en/latest/index.html) in its default configuration -- it'll execute notebooks if they're committed without any output. And since the `model-forced.ipynb` had its outputs cleared, it's now trying to execute it to generate it fresh. I agree that'd be the way to do it! I guess we just need to look into what kernels are available in the RTD environment so it can actually run it properly...
> > 
> > 
> > Ah yeah that's my bad - maybe we should just re-run the notebooks then commit? I deleted the output since it's a bit messy but this could be cleaned up. e.g we don't need the dask dashboard and other clutter. Currently the only useful output in the notebooks is the plot of the topography anyway
> 
> Nah.. Let's figure out how to actually have the notebooks run by the docs. That way the docs will be up to date by construction!

Sure but I don't know how we'll be downloading and interpolating the gebco dataset on a cloud based machine?

regional_mom6/regional_mom6.py
```diff
@@ -934,31 +957,70 @@ def bathymetry(
...
+
+            try:
+                regridder = xe.Regridder(bathyout, tgrid, "bilinear", parallel=parallel)
+            except TypeError:
+                print(
+                    "Warning: xesmf parallel regridding failed. You likely have an older version of xesmf installed - try updating to 0.8.x or higher."
                 )
+                bathyout = bathyout.chunk({"lon": -1, "lat": -1})
+                tgrid = tgrid.chunk({"lon": -1, "lat": -1})
+                regridder = xe.Regridder(bathyout, tgrid, "bilinear")
```
ashjbarnes, line 1015:
```suggestion
```

ashjbarnes, line 1015:
```suggestion
            regridder = xe.Regridder(bathyout, tgrid, "bilinear", parallel=parallel)
```

README.md
ashjbarnes, line 68:
```suggestion
```

regional_mom6/regional_mom6.py
navidcy: I don't understand why this will need testing, not that it's of course always good to have tests.

regional_mom6/regional_mom6.py
navidcy: add a comment?

navidcy:
> Sure but I don't know how we'll be downloading and interpolating the gebco dataset on a cloud based machine?

I presume the "don't know" bit goes to the downloading, right?
Interpolating is only a matter of making the example simple enough so that is computationally feasible. Am I right?

regional_mom6/regional_mom6.py
ashjbarnes: This passes for gebco, but this function is also used for slicing surface forcing and the initial condition. In this implementation we're asserting the resolution that this function will work with. For example it fails if you use it on access om2 01, breaking the `model forced` notebook

ashjbarnes:
> > Sure but I don't know how we'll be downloading and interpolating the gebco dataset on a cloud based machine?
> 
> I presume the "don't know" bit goes to the downloading, right? Interpolating is only a matter of making the example simple enough so that is computationally feasible. Am I right?

Yeah that's right. So this would mean re-doing the example notebooks to only download sub samples of GEBCO through some API. I've already done this for GLORYS, but often you have to re-run the bash command because the server will hang or reject your request even for small dataset. This would mean our tests rely on an external server, adding a bit more complexity. This would be a lot of work and also deserving of its own PR I think

navidcy:
Fair point

why don’t we rerun the notebooks so that all cells are evaluated and open an issue to remind us to convert the notebook examples to ones evaluated on the spot by the docs. How does this sound?

regional_mom6/regional_mom6.py
navidcy: You are saying that a function called `bathymetry` is used to process surface forcing?

ashjbarnes: Ohh no you're right this is the code that's calling nicer slicer, my bad sorry. But the point still stands that we are making an assumption about the bathymetry file. If we're happy that no-one would use bathymetry that's coarser than 10th degree then this does pass for Gebco

navidcy: Ok. But still you are saying that access-on2 bathymetry does not satisfy

```python 
np.isclose(bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][0], 360, atol=1e-2)
```

I find this puzzling. How much is

```python 
bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][0]
```
?

navidcy: I’m actually in disagreement with you in the sense that the code you suggest is making an assumption while what I suggest does not. If what I suggest makes an assumption the can you please explain me what the assumption is? If what we want is whether the bathymetry goes round the globe in longitude then I feel the least assumptions come from ensuring the longitude extent is 360 degrees.

navidcy: However, even if I'm not convinced, I'm not gonna step my foot down. Happy to keep it as you had it and I'll raise an issue to remind us to fix it?

For this PR we should definitely fix the dependencies-environment related issues, e.g., https://github.com/COSIMA/regional-mom6/pull/73/files#r1325511524

navidcy:
I think the PR title should be updated as this PR does a bit more than that, right?

regional_mom6/regional_mom6.py
ashjbarnes: If we change the tolerance to `1e-1` then it's ok for access om2 01. This is the longitude extent for it:

image

How about we change the tolerance to 0.1 (any resolution less than this would be silly for a regional model anyway) whereas `1e-2` would fail for other datasets that were coarser for 100th of a degree. It works for Gebco because it wraps around with a common start and end point of 180 / -180

I agree though that we should focus on getting these bigger fixes right in this PR and should do some nice quality of life PRs later once package is functioning and executing the example notebooks

ashjbarnes:
> Fair point
> 
> why don’t we rerun the notebooks so that all cells are evaluated and open an issue to remind us to convert the notebook examples to ones evaluated on the spot by the docs. How does this sound?

@micaeljtoliveira suggested to me yesterday that we use a docker image for testing that has pre-uploaded data for interpolation. This could be really handy for evaluating an example notebook (after it's modified to accommodate)! We could do a separate PR that replaces the ubuntu-latest docker images with a custom one that includes compiled FRE tools as well as some trimmed down bathy / boundary / IC data. @angus-g  do you have thoughts on using a custom docker image for the testing env?

angus-g:
> do you have thoughts on using a custom docker image for the testing env?

Yeah, I think that's a good idea.

ashjbarnes:
> > do you have thoughts on using a custom docker image for the testing env?
> 
> Yeah, I think that's a good idea. If you have some of the testing data to point me to, I can build and set that up.

Here are some cut outs I generated from the example notebook: `/scratch/v45/ab8992/regional_tmp/tasmania-september20-23` 

If all of those files plus `/g/data/ik11/inputs/JRA-55/RYF/v1-3/*.1990_1991.nc` would be sufficient. We could also cut down the JRA forcing to `lat: [-50 : -37 , lon 150 : 155]` to cut down on storage space

What would make this perfect is if there was also a compiled MOM6 executable and FRE tools ready to go. That way part of the testing suite would also be running MOM6 for one day on the output

ashjbarnes:
it was a long journey but we finally got there 
image

Just evaluated the notebooks as suggested so that sorted out the docs issue. In future agree that we should have one main example notebook that can be evaluated by tests via a docker image

regional_mom6/regional_mom6.py
navidcy: How about:

```suggestion
            if np.isclose(bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][0] - horizontal_resolution, 360)
```

or whatever variable name the horizontal resolution is?

What you posted above suggests that I was forgetting that the two ends of a periodic grid are not exactly the same point (since that would be redundant) but they differ by a `Δx`.

navidcy:
> it was a long journey but we finally got there image
> 
> Just evaluated the notebooks as suggested so that sorted out the docs issue. In future agree that we should have one main example notebook that can be evaluated by tests via a docker image

Great! I'm happy to leave the docker-tests for another PR.
But we should deal with this bottleneck dependency in this PR.

regional_mom6/regional_mom6.py
navidcy: Why is this bathymetry on `xu_ocean` tho? Bathymetry lives on cell centers. So even in access-om2 I'd expect

```python
bathymetry['xt_ocean'][0] = -279.95
bathymetry['xt_ocean'][-1] = 80.05
```

right?

ashjbarnes:
Is there much I can do to help with setting up the environment correctly (i.e elegantly resolving the bottleneck issue). This is become a massive thread, but am I right in understanding this is the last thing we'd like to sort out before merging?

In the meantime is there an easy way to flag to would-be users that they should use this branch instead of the currently-not-working main branch? Temporary change to README is a bit clunky but could work

angus-g:
I think I lost track of what's going on here. Which environment are you referring to, and what is the bottleneck issue?

navidcy:
> Is there much I can do to help with setting up the environment correctly (i.e elegantly resolving the bottleneck issue). This is become a massive thread, but am I right in understanding this is the last thing we'd like to sort out before merging?

@angus-g can you help here?

> In the meantime is there an easy way to flag to would-be users that they should use this branch instead of the currently-not-working main branch? Temporary change to README is a bit clunky but could work

Telling users to use a branch is bad practice.
Issue #72 does alarms users that there is something wrong and this PR will sort it out? (They are linked.)

environment-ci.yml
navidcy: @angus-g this bottleneck here

regional_mom6/regional_mom6.py
ashjbarnes: Yeah that would work. You'd need to calculate the horizontal resolution for the given input bathymetry first via `bathy[varnames["xh"]][1] - bathy[varnames["xh"]][0]` or something since that's not stored

navidcy:
> I think I lost track of what's going on here. Which environment are you referring to, and what is the bottleneck issue?

https://github.com/COSIMA/regional-mom6/blob/1e86b7de1ef9460ea490d46a217ebd98ba998f30/environment-ci.yml#L6

regional_mom6/regional_mom6.py
navidcy:
```python
if np.isclose(bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][1], 360)
```
?

ashjbarnes:
> I think I lost track of what's going on here. Which environment are you referring to, and what is the bottleneck issue?

Ok here's a rough summary in chronological order

1. This PR was made because when users follow the installation instructions currently in main branch, the package doesn't work, and never has because of issues with the XESMF install being janky.  
2. In patching this, I found that using the latest version of XESMF allows for a parallel implementation in Python, bypassing the need for the janky subprocess calls I'd ben using for regridding the topography. **This fixed the original issue**
3. BUT, there was a further issue with the environment surrounding the [bottlneck](https://github.com/COSIMA/regional-mom6/pull/73#issuecomment-1718624950) package required by `.bfill`. This may have been caused by my changing the `environment-ci.yml` file or may have been an underlying issue that never arose because it used to break at the regrid step earlier.
4. **Current state:** Functions I wrote that test the whole regridding pipeline work, but Angus indicated that he'd like to check that I did the environment stuff right, which is more than fair enough since this was mostly trial end error on my part

regional_mom6/regional_mom6.py
navidcy: Or even cleaner

```python

horizontal_resolution = bathy[varnames["xh"]][1] - bathy[varnames["xh"]][0]

if np.isclose(bathy[varnames["xh"]][-1] - bathy[varnames["xh"]][0] - horizontal_resolution, 360)
```

navidcy:
@ashjbarnes what "extra 8" is this comment referring to?
https://github.com/COSIMA/regional-mom6/blob/d9b997f0e497d58db4a827432be7c88d2b472b1b/regional_mom6/regional_mom6.py#L123C20-L123C53

README.md
navidcy: @ashjbarnes we should deal with this also; please remove all NCI refs from README?

regional_mom6/regional_mom6.py
navidcy: @ashjbarnes ?

tests/test_expt_class.py
navidcy: cc @ashjbarnes

ashjbarnes:
> @ashjbarnes what "extra 8" is this comment referring to?
> 
> https://github.com/COSIMA/regional-mom6/blob/d9b997f0e497d58db4a827432be7c88d2b472b1b/regional_mom6/regional_mom6.py#L123

Idk probably a typo. This part of the code hasn't been changed in a long time

ashjbarnes:
With the change of the dependency files you made @angus-g it seems like the tests still go through ok. Is there anything else that needs to be checked to make sure the installation is working as expected?

angus-g:
> it seems like the tests still go through ok. Is there anything else that needs to be checked to make sure the installation is working as expected?

Does it still work for your environment creation on Gadi, etc.?

ashjbarnes:
> > it seems like the tests still go through ok. Is there anything else that needs to be checked to make sure the installation is working as expected?
> 
> Does it still work for your environment creation on Gadi, etc.?

I haven't been able to figure out how to get my Gadi-Jupyter notebook kernel to use recognise custom environments. It's very hit and miss! I'll run the test scripts on Gadi though even if not the example notebooks

angus-g:
Have you used the [IPython instructions](https://ipython.readthedocs.io/en/latest/install/kernel_install.html#kernels-for-different-environments) for letting the notebooks know about your different environments?

ashjbarnes:
> Have you used the [IPython instructions](https://ipython.readthedocs.io/en/latest/install/kernel_install.html#kernels-for-different-environments) for letting the notebooks know about your different environments?

Yeah I've set up a new conda env twice now in my Gdata (exceeds quota in home), and this second time I've renamed it as in the instructions you link here. Path to gdata conda is also in my `$PATH`. However jupyter doesn't detect it 

image


I've run the tests on Gadi at least, so am confident that the regridding works now after following install instructions, 
even if the example notebooks haven't been tested outside of loading the functions within the analysis3 environment

angus-g:
> This isn't strictly a dependency of the package but everyone using it will want matplotlib to see their domain. Should we include it in the dependencies file?

No, it should remain optional. Suppose somebody wanted to run in batch mode without plotting, they wouldn't want/need matplotlib, and it could complicate headless installs.

> Likewise since Bokeh is missing you get this message when trying to view the Dask cluster. Should we include that too?

Definitely not, that's much more dependent on one's environment setup. The way to do this is to provide a conda environment with all the correct packages installed, but that can be difficult to generalise. I don't think providing a working dask cluster setup is in the scope of this project.

ashjbarnes:
> > This isn't strictly a dependency of the package but everyone using it will want matplotlib to see their domain. Should we include it in the dependencies file?
> 
> No, it should remain optional. Suppose somebody wanted to run in batch mode without plotting, they wouldn't want/need matplotlib, and it could complicate headless installs.
> 
> > Likewise since Bokeh is missing you get this message when trying to view the Dask cluster. Should we include that too?
> 
> Definitely not, that's much more dependent on one's environment setup. The way to do this is to provide a conda environment with all the correct packages installed, but that can be difficult to generalise. I don't think providing a working dask cluster setup is in the scope of this project.

Yes this all makes sense! In that case what are the remaining actionable things in this PR?

navidcy:
> > > This isn't strictly a dependency of the package but everyone using it will want matplotlib to see their domain. Should we include it in the dependencies file?
> > 
> > 
> > No, it should remain optional. Suppose somebody wanted to run in batch mode without plotting, they wouldn't want/need matplotlib, and it could complicate headless installs.
> > > Likewise since Bokeh is missing you get this message when trying to view the Dask cluster. Should we include that too?
> > 
> > 
> > Definitely not, that's much more dependent on one's environment setup. The way to do this is to provide a conda environment with all the correct packages installed, but that can be difficult to generalise. I don't think providing a working dask cluster setup is in the scope of this project.
> 
> Yes this all makes sense! In that case what are the remaining actionable things in this PR?

This: https://github.com/COSIMA/regional-mom6/pull/73#discussion_r1338043000 ?

README.md
navidcy: f09f566e88fa95360b395842041f453cde04c2ae deals with this

navidcy:
I pushed my suggestion for the check whether bathymetry goes all along the domain. Let me know if you have objection still on this @ashjbarnes.

Let's see if tests pass. I'm happy at this point. I'll approve.

navidcy (approved):

navidcy:
I think @angus-g also need to re-review/approve because of the requested changes.

navidcy:
oh yeah, I had forgotten the temp dir fixtures issue! thanks @angus-g!

angus-g (approved):
I think it's good to go now.

PR #75, opened 2023-09-28
Use Path objects instead of string manipulation

angus-g:
Importantly, reduces number of subprocess/os calls, in favour of using Python-native operations.


navidcy (approved):
seems cleaner

navidcy:
@angus-g is this done? feel free to merge :)

Issue #52, opened 2023-06-22
Docstring regarding the resolution in latitudinal directions is slightly ambiguous


navidcy:
This does not read right to me:

https://github.com/COSIMA/regional-mom6/blob/abf630d3fa909f0e609eaba167d454419f96e812/regional_mom6/regional_mom6.py#L509

`yextend` is a tuple with the extreme latitude values. But if these are symmetric about the equator then  `mean(yextent) = 0` and `res_y = 0`?

ashjbarnes:
res_y is a modifier of latitude cell widths. At the equator this should be 1, since variations in lat and lon represent the same spatial extent. As you move to the poles you want this to decrease. I think what you're describing is the intended behaviour given that cos(0) = 1

ashjbarnes:
What's the status of this issue? Is there a proposed change?

navidcy:
I think I understand what `res_y` means to be and indeed it is also explained in the docstring.

However, my problem here perhaps points to the docstring which seems a bit contradictory.

https://github.com/COSIMA/regional-mom6/blob/d5326504c210cc328080e42028e3100e45066695/regional_mom6/regional_mom6.py#L521-L532

1. The `_make_hgrid` function as defined here doesn't seem to have a default behavior.
2. "This way, grids in the centre of the  domain are perfectly square" is wrong as they are not squares but rather spherical quadrilateral with equal Δlongitude and Δlatitude. Shall we rephrase?
3. 2. "latitude spaced with 0.1 cos(mean_latitude)" so that the grid cells at the center of the domain are squares... But that implies that the latitude grid is centered around the equator, right? What if the latitude domain is `[-80, -60]`. Then the center cell will have `Δlongitude = 0.1 deg` and `Δlatitude = 0.1 * cos(70degrees)` so they are not "squares".

ashjbarnes:
Ohh true - that docstring is out of date! Good spotting. That would have been when _make_hgrid was using the rectangular hgrid function. I then modified it to be a wrapper for whatever grid gen scheme the user chooses (of which we currently have one). I think altering this docstring will fix this ambiguity

navidcy:
Wanna give it a go?

PR #82, opened 2023-11-15
Update `_make_hgrid` docstring

ashjbarnes:
Closes issue #52

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -510,7 +510,7 @@ class experiment:
     def _make_hgrid(self, gridtype):
         """Sets up hgrid based on users specification of
         domain. Default behaviour leaves latitude and longitude evenly
-        spaced. This is very simple but suitable for small domains. 
+        spaced. This is very simple but suitable for small domains.
 
         Note:
             The intention is for the hgrid generation to be very flexible. In future updates, further options will built
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -513,9 +513,7 @@ class experiment:
         spaced. This is very simple but suitable for small domains.
 
         Note:
-            The intention is for the hgrid generation to be very flexible. In future updates, further options will built
-            into _make_grid(), probably wrapping some functions from the `Grid Gen repository <https://github.com/ESMG/gridtools>`_.
-            For now, you can customise the hgrid by simply overwriting the `hgrid.nc` file that's deposited in your `rundir` after initialising an `experiment`. To conserve the metadata, it might be easiest to read the file in, then modify the fields before re-saving.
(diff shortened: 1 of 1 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -510,14 +510,12 @@ def __init__(
...
-        spaced this way and latitude spaced with 0.1
-        cos(mean_latitude). This way, grids in the centre of the
-        domain are perfectly square, but decrease monatonically in
-        area away from the equator
+        spaced. This is very simple but suitable for small domains.
 
+        Note:
+            The intention is for the hgrid generation to be very flexible. In future updates, further options will built
+            into _make_grid(), probably wrapping some functions from the `Grid Gen repository <https://github.com/ESMG/gridtools>`_.
+            For now, you can customise the hgrid by simply overwriting the `hgrid.nc` file that's deposited in your `rundir` after initialising an `experiment`. To conserve the metadata, it might be easiest to read the file in, then modify the fields before re-saving.
```
navidcy, line 518: no need to promise things for the future, let's leave only what can happen now... when this future time come that we have implemented more things we will include it in the docs, right?

ashjbarnes: I think it's still important to tell users how to modify the respective grids. The intention of the package isn't to prescribe my two example grids, but leave things flexible. I agree making promises for the future is a bit odd but in this case I thought it gave an explanation of the intention of this function

navidcy: Keep that bit! But don't make promises for how this package will be in the future.

ashjbarnes: done

regional_mom6/regional_mom6.py
```diff
@@ -510,14 +510,10 @@ def __init__(
     def _make_hgrid(self, gridtype):
         """Sets up hgrid based on users specification of
         domain. Default behaviour leaves latitude and longitude evenly
-        spaced.
-
-        If user specifies a resolution of 0.1 degrees, longitude is
-        spaced this way and latitude spaced with 0.1
-        cos(mean_latitude). This way, grids in the centre of the
-        domain are perfectly square, but decrease monatonically in
-        area away from the equator
```
navidcy, line 519: why was this info deleted?
this is what happens with `gridtype == "even_spacing"`?
will we loose this information or is it somewhere else?

ashjbarnes:
this is the part of the docstring that you didn't like in issue #52, hence why I was deleting it to try and resolve this issue

regional_mom6/regional_mom6.py
ashjbarnes: this is the part of the docstring that you didn't like in issue https://github.com/COSIMA/regional-mom6/issues/52, hence why I was deleting it to try and resolve this issue

navidcy: Hahahha

navidcy: Apparently I forgot

navidcy (approved):

Issue #79, opened 2023-11-09
Setting up Regional-MOM6 on non-NCI system


croachutas:
Hi,
I'm in the process of setting up Regional-MOM6 on my workstation and working through the reanalysis forcing example as a bit of an exercise to get up to speed on it before using it as a teaching aid for one of the IMAS-OUC 2+2 programme units.

Anyway, I've thus far encountered a few issues:
1. motu-client isn't listed as a dependency, easy to fix at my end (just pip...) but you probably should add it to the project.toml so this doesn't catch people out in future.

2. toolpath points to MOM5 tools instead of tools available with MOM6 or independently:
`toolpath = "/home/157/ahg157/repos/mom5/src/tools/" `
Again, this is something I can work around (download and build MOM5...), but I think this dependency on MOM5 needs to be clearly stated. Ideally either alternative tools available with MOM6 should be used or the necessary tools should be provided with Regional-MOM6.

3. Running the code to download GLORY boundary forcing and initial states successfully creates get_oceanfiles.sh but when the code tries to run the shell script I get warnings to the effect of: `/home/croach/anaconda3/envs/MOM6_tools/bin/python: No module named motuclient.__main__; 'motuclient' is a package and cannot be directly executed` and no data is downloaded. Looking at https://github.com/clstoulouse/motu-client-python/tree/master#Usage and comparing that to the contents of get_oceanfiles.sh I suspect that rm.motu_requests is written for motu-client v1.8.0 instead of motu-client v3.x. I can probably downgrade to motu-client v1.8.0 easily enough.

Anyway, I'll have a crack at installing the MOM5 tools and downgrading motu-client and keep you updated on if that fixes my current problems.

-Chris Roach

croachutas:
Okay, downgrading to motu-client v1.8.0...
`/home/croach/anaconda3/envs/MOM6_tools/lib/python3.11/site-packages/motuclient.py:31: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
  from cgi import log
2023-11-09 13:14:46.802 [ INFO] Asynchronous mode set
2023-11-09 13:14:46.803 [ INFO] Authenticating user croach3 for service https://my.cmems-du.eu/motu-web/Motu
2023-11-09 13:14:47.189 [ERROR] Execution failed: <urlopen error [SSL] internal error (_ssl.c:1006)>
[...]

Issue #59, opened 2023-06-23
Convert demo that uses ACCESS-OM2 output to `cosima-recipes` example


navidcy:
At the moment there are two demos: one using GLORYS reanalysis and one using ACCESS-OM2 output.

Given that the ACCESS-OM2 output is not available to anyone outside Australia's HPC I suggest that at some point we move the ACCESS-OM2 demo into [cosima-recipes](https://github.com/COSIMA/cosima-recipes) (e.g. as part of https://github.com/COSIMA/cosima-recipes/pull/216?) and leave inside the package the demo that can be, in principle, executed by anyone.

We can always (and we should!) add a link to the cosima-recipes example in the package's docs.

ashjbarnes:
I agree with this!

navidcy:
I'm closing this since after #162 it's not related to this repo but it's left with cosima-recipes to deal with ;)

navidcy:
> I'm closing this since after #162 it's not related to this repo but it's left with cosima-recipes to deal with ;)

@luweiyang, what I mean here is that #162 deleted this example from this repository and then https://github.com/COSIMA/cosima-recipes/pull/338 when merged, will move the ACCESS-OM2 example in the cosima-recipes repo

PR #83, opened 2023-11-15
Declutter example notebook

ashjbarnes:
Here I've removed all descriptions of Gadi / NCI and almost all of Payu from the `reanalysis_forced` notebook, and deleted the model forced one. This will be migrated to cosima recipes

The old notebook cells for "setup era5" and "set up run directory" are now methods of the experiment class which further streamlines the notebook. The `using_payu` flag is included here to give users the option of a `config.yaml` file being generated for their experiment. It defaults to False. 

The default input directory has also been decluttered to make things generic and not Gadi-centric

Finally, to address the deprecation of the motu client (issue #79 ) I've removed this entirely and given users step by step instructions for using the GUI on Copernicus's website to download the required data.

Related to #59 in that I've updated the model-forced example but I won't delete it until we find a good home for it elsewhere (see https://github.com/COSIMA/cosima-recipes/issues/308 )


ashjbarnes:
I'm going to update / fix the 'model forced' one then re-open this PR. Until we find a landing place for it I'll at least iron out some of the issues pointed out by Luwei

ashjbarnes:
I've re-added the ACCESS-OM2 example but made it clear in the markdown that it's intended for NCI users. This should be migrated elsewhere in COSIMA when we agree on the space

ashjbarnes:
Codecov is down because I've moved the `process era5` functionality to the package. A future PR that expands testing should address this rather than me making this PR longer. I've manually tested both notebooks and run the corresponding MOM6 experiments they produce

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+
+        ## Get mask table information
+        ncpus = 10
+        mask_table = None
+        for i in os.listdir(f"{self.mom_input_dir}"):
+            if "mask_table" in i:
+                mask_table = i
+                a = mask_table.split(".")[1]
+                b = mask_table.split(".")[2].split("x")
+                ncpus = int(b[0]) * int(b[1]) - int(a)
```
angus-g, line 1260: An alternative could be to parse from the mask file directly (the filename could lie!)

```python
mask_tables = self.mom_input_dir.glob("mask_table.*")
for mask_table in mask_tables:
  with open(mask_table, "r") as f:
    masked_cpus = int(f.readline())
    x, y = (int(v) for v in f.readline().split(", "))
```

Otherwise, at least:
```suggestion
        for p in self.mom_input_dir.glob("mask_table.*"):
          mask_table, masked, layout = p.name.split(".")
          x, y = (int(v) for v in layout.split("x"))
          ncpus = (x * y) - masked
```

But, what if there are multiple mask tables? This (both mine and yours) will default to the lexicographically last. Also, the layout we're going to use is set by `FRE_tools()`, but the core count is determined here, at run time. If a mask file does exist, but `FRE_tools()` wasn't called, `self.layout` won't exist and you'll hit an error later on. If a different mask file exists to the one `FRE_tools()` creates, the core count and layout could be out of sync.

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+        if mask_table == None:
+            print("No mask table found! Run FRE_tools first. Terminating")
+            raise ValueError
```
angus-g, line 1263:
```python
for p in self.mom_input_dir.glob("mask_table.*"):
  ...
else:
  raise ValueError("No mask table found. Run FRE_tools() method first to generate one.")
```

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+
+        print("Number of CPUs required: ", ncpus)
```
angus-g, line 1265: Necessary?

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+
+        ## Modify MOM_input
+        inputfile = open(f"{self.mom_run_dir}/MOM_input", "r")
```
angus-g, line 1268: Instead of all this string manipulation, you could modify `input.nml`, such as:

```nml
&MOM_input_nml
	output_directory = '.',
	input_filename = 'n',
	restart_input_dir = 'INPUT',
	restart_output_dir = 'RESTART',
	parameter_filename = 'MOM_input', 'MOM_override', 'MOM_layout'
/
```

Then write to `MOM_layout` directly.

```python
layout_params = {
  "MASKTABLE": mask_table,
  "LAYOUT": f"{self.layout[0]}, {self.layout[1]}",
  ...
}

with open(self.mom_input_dir / "MOM_layout", "w") as f:
  for param, val in layout_params.items():
    f.write(f"{param} = {val}\n")
```

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+                # lines[i] = f"NJGLOBAL = {str(y_indices_centre[1] - y_indices_centre[0])}\n"
+                lines[i] = f"NJGLOBAL = {self.hgrid.ny.shape[0]//2}\n"
+
+        inputfile = open(f"{self.mom_run_dir}/SIS_input", "w")
+        inputfile.writelines(lines)
+        inputfile.close()
+
+        ## If using payu to run the model, create a payu configuration file
+        if not using_payu:
+            os.remove(f"{self.mom_run_dir}/config.yaml")
```
angus-g, line 1315: This doesn't really need to be *removed* in the non-payu case, does it?

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+            for i in range(len(lines)):
+                if "ncpus" in lines[i]:
+                    lines[i] = f"ncpus: {str(ncpus)}\n"
+
+                if "input:" in lines[i]:
+                    lines[i + 1] = f"    - {self.mom_input_dir}\n"
+
+            inputfile = open(f"{self.mom_run_dir}/config.yaml", "w")
+            inputfile.writelines(lines)
+            inputfile.close()
```
angus-g, line 1331:
```python
with open(self.mom_run_dir / "config.yaml", "r") as f:
  payu_config = yaml.safe_load(f)

payu_config["ncpus"] = ncpus
payu_config["input"].append(self.mom_input_dir)

with open(self.mom_run_dir / "config.yaml", "w") as f:
  yaml.dump(payu_config)
```

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+            for i in range(len(lines)):
+                if "current_date" in lines[i]:
+                    tmp = self.daterange[0]
+                    lines[
+                        i
+                    ] = f"{lines[i].split(' = ')[0]} = {int(tmp.year)},{int(tmp.month)},{int(tmp.day)},0,0,0,\n"
+
+            inputfile = open(f"{self.mom_run_dir}/input.nml", "w")
+            inputfile.writelines(lines)
+            inputfile.close()
```
angus-g, line 1346: As with the yaml example, this should be done using Marshall's https://github.com/marshallward/f90nml

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+                        * (
+                            10
+                            ** (
+                                8.07131
+                                - 1730.63 / (233.426 + rawdata["2d"]["d2m"] - 273.15)
+                            )
+                        )
+                        * 101325
+                        / 760
+                    }
```
angus-g, line 1411: Because we're so deeply nested, Black kind of mangles this... Maybe split out the exponent, which reformats much more nicely (naming likely incorrect, too many magic constants for me to really know what's going on here):

```python
                dewpoint = 8.07131 - 1730.63 / (233.426 + rawdata["2d"]["d2m"] - 273.15)
                humidity = (
                    (0.622 / rawdata["sp"]["sp"]) * (10**dewpoint) * 101325 / 760
                )
```

angus-g (changes requested):

regional_mom6/regional_mom6.py
ashjbarnes: Oh yeah good idea! I know this code is a mess. I've been kind of developing as I go with my own research but haven't wanted to get too distracted on `regional-mom6` stuff so didn't do too much prettying-up for this PR

regional_mom6/regional_mom6.py
ashjbarnes: No, I was just trying to stick with the mantra of not being NCI-centric. Could be confusing to other users if they have a random config file that doesn't work for their system right?

regional_mom6/regional_mom6.py
ashjbarnes: Thanks! Would this repo then be included as a dependency?

angus-g: Yes, but it's a pretty low overhead dependency.

regional_mom6/regional_mom6.py
angus-g: I guess so, but it doesn't seem much more intrusive than all the other random files in the config directory.

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,204 @@ def FRE_tools(self, layout):
...
+            using_payu (Optional[bool]): Whether or not to use payu to run the model. If True, a payu configuration file will be created.
+
+        """
+
+        ## Copy the default directory to the run directory
+
+        subprocess.run(
+            f"cp {str(Path(rmom6_path) / 'regional_mom6' / 'default_rundir' / surface_forcing)}_surface/* {str(self.mom_run_dir)}",
+            shell=True,
+        )
```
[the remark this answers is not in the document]
ashjbarnes: I remember I did try using copytree here but it doesn't support the overwriting of files, so I went back to a subprocess. What's the proper way to copy and force overwrite? `distutils`?

regional_mom6/regional_mom6.py
ashjbarnes: Resolved by just handling the 'no mask table' case properly

regional_mom6/regional_mom6.py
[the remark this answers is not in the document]
angus-g: You can do `shutil.copytree(src, dst, dirs_exist_ok=True)`

regional_mom6/regional_mom6.py
ashjbarnes: Yeah I'm not sure how to handle all the possibilities here. The point of this function is to really hold the hand of a new user. I'll get it to print a warning if there are multiple mask tables to prompt them to delete or move unwanted ones

regional_mom6/regional_mom6.py
ashjbarnes: Which files in the config  directory are you referring to? I've tried to tidy it up now to leave only the essentials. I think `config.yaml` would now be the only one that's not MOM specific

angus-g: I suppose I mean that the MOM-specific ones are pretty messy already, but I'm not fussed if you want to remove it.

regional_mom6/regional_mom6.py
ashjbarnes: Done

regional_mom6/regional_mom6.py
```diff
@@ -1222,6 +1225,207 @@ def FRE_tools(self, layout):
...
+        rundir_src = (
+          Path(__file__).parent /
+          "default_rundir" /
+          f"{use_default}_surface"
+        )
+        shutil.copytree(rundir_src, self.mom_run_dir, dirs_exist_ok=True)
+        subprocess.run(
+            f"cp {str(rundir_src)}/* {str(self.mom_run_dir)}",
+            shell=True,
+        )
```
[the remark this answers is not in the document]
ashjbarnes: Thanks Angus, I did this then went back because there were merge conflicts due to committing from here and my copy at the same time

regional_mom6/regional_mom6.py
ashjbarnes: Where do we include these new dependencies? 

I'm a bit confused with the structure. Which things go in "pyproject.toml" and which go in "environment-ci.yml"? Is esmpy confined to the environment folder because its a conda rather than pip install?

In that case I'll just add `pyyaml` to `environment-ci.yml`

angus-g: It should go in the pyproject.toml, because they're the Python dependencies, not the one specific to conda.

regional_mom6/regional_mom6.py
ashjbarnes: Similar to comment above, does this just go in "dependencies" field of the pyproject.toml file?

angus-g: Yes.

regional_mom6/regional_mom6.py
angus-g: Also, I think it should be called `PyYAML` (although I don't know if capitalisation matters for packages...)

regional_mom6/regional_mom6.py
ashjbarnes: This would break if the `MOM_input` file already had parameter stuff right? Because otherwise we'd need to do an override?

Say if user would BYO a run directory and then opt to modify it, they'd likely already have layout parameters and then you'd get duplicates. Maybe it's just easier to modify `MOM_input` or override the layout?

regional_mom6/regional_mom6.py
ashjbarnes: Feedback from @navidcy was to remove nci specific things. I don't mind either except that it's probably easier for others to understand what's going on if there aren't unfamiliar non-MOM files floating around

regional_mom6/regional_mom6.py
ashjbarnes: pyyaml mangles the manipulation of `config.yaml` for some reason. pyyaml fails to read in the multiple input lines (just gets the first one) and then re-writes the yaml file in alphabetical order.

I don't have time to fiddle with this so I'm going to revert back to the string manipulations

ashjbarnes: Latest commit reverts to string manipulations and has been tested on example notebooks.

regional_mom6/regional_mom6.py
ashjbarnes: I've stuck to just modifying the `MOM_input` file since I think this is ultimately more user friendly, and then we don't have to worry about existing / not existing layout parameters in `MOM_input`

regional_mom6/regional_mom6.py
ashjbarnes: Thanks Micael but again this isn't a priority for now. I've already given it some time, and it didn't work. If we want to update / tidy the existing code then that can be a separate PR.

ashjbarnes: @micaeljtoliveira Thanks, yes would be good to do one day! Sorry for the short response - I feel sometimes that there are a million things to do with this repo and I'm always tempted to keep working away instead of other priorities with my PhD. Hence I'm trying to be careful about keeping PRs on task and not getting too sidetracked

ashjbarnes:
Are we happy to tidy up string manipulations in a separate PR later? For now things are working even if a bit ugly for the `yaml` editing

angus-g:
Apologies, fixed. I had no clue you were working live off a PR branch, I would strongly advise against doing that.

ashjbarnes:
If not, then from a github workflow perspective what's the best way for me to keep working for the moment? I'm losing a lot of time trying to juggle the codebase when this PR is unresolved and I'm already working on the next one (and need it for my day-to-day research). It was becoming too much admin to continually merge changes from this branch into the `withtides` branch, especially when merge conflicts cropped up. 

I then manually made a new branch forked from `declutter_notebook` to build off but then had to troubleshoot why things weren't working. Perhaps there's a better way of doing things that I'm not aware of?

navidcy:
I haven't followed any discussion here. @ashjbarnes, @angus-g is this PR under control? If you need help ping me.

ashjbarnes:
Latest commit implements what we talked about today @angus-g . Notebook has been tested but I'll go back and modify notebooks to add some more documentation - that'll be the last planned commit from me unless something else comes up / breaks!

Thanks @navidcy ! Almost finished :)

regional_mom6/regional_mom6.py
ashjbarnes: This is useful for non-payu users as they'll need to enter in the number of cpus into whatever system they're using

angus-g (approved):

regional_mom6/regional_mom6.py
```diff
@@ -585,9 +590,15 @@ def ocean_forcing(
...
 
-        if varnames["time"] in ic_raw.dims:
+        if 'time' in ic_raw.variables:
+            ic_raw = ic_raw.drop_vars('time')
+        elif varnames["time"] in ic_raw.dims:
             ic_raw = ic_raw.isel({varnames["time"]: 0})
 
+
+
+        if varnames["time"] in ic_raw
```
navidcy, line 600: this can't be right... python can't even parse it

navidcy (dismissed):

regional_mom6/regional_mom6.py
ashjbarnes: Don't worry about this, I was testing this code one last time after my patches and bugfixing. I was just using github to sync my local changes to gadi so I could run the example notebooks. I'll finish testing, then commit the last fixes and then merge

navidcy:
tip: 
before pushing you should format with black otherwise tests don't run... and tests that fail might hint you about something wrong that is happening. If you postpone formatting black (and thus running the tests) you may not see a bug you introduced until way after you've put substantial effort in refactoring something that you may need to undo.

regional_mom6/regional_mom6.py
```diff
@@ -11,9 +11,13 @@
 import netCDF4
 from dask.distributed import Client, worker_client
 from dask.diagnostics import ProgressBar
+import f90nml
 import datetime as dt
 import warnings
+import shutil
+import os
 from .utils import vecdot
+import yaml
```
navidcy, line 20: did we add these packages as dependencies?

navidcy: I see only 

```python
  "PyYAML>=6.0.1",
  "f90nml>=1.4.1",
```

what about `os` and `shutil`?

angus-g: They're both part of Python itself.

ashjbarnes:
Everything tested and working with latest executable. Ready to merge

PR #90, opened 2024-01-18
Update README.md to specify MOM6 version requirements and configuration

### LINKED ISSUE #87: Specify required versions of MOM6, FMS & FMScoupler
The core library in this package will work with the versions of MOM6 & FMS code specified in [MOM6_examples](https://github.com/NOAA-GFDL/MOM6-examples) as of Jan 2024. However, the code here is quite old and doesn't include the latest improvements to open boundary conditions, or to surface fluxes (required to use ERA5 forcing). This was made apparent by issues Chris found in issue #79 

Things fall over when running our examples though, as the `MOM_input` and `input.nml` files need minor changes to reflect versions used. 

Need to link specific commits to all three of these codebases, and add some rudimentary instructions for how someone modifies the `MOM_examples` repo to include updated code. Compiling is another story though and probably beyond the scope of our documentation

ashjbarnes:
closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ or
 pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de0566df329c4ea472002f7e
 ```
 ## MOM6 Configuration and Version Requirements
-While the underlying package should work for coupled or ocean only configurations, the premade files for the demos are written for a coupled MOM6-SIS2 configuration.
+The package and demos were written for a coupled SIS2-MOM6 configuration, and may also work for an ocean only MOM6 run but this has not been tested. 
 
-The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. Note that GFDL's `MOM6_examples` repository uses older versions of the Flexible Modelling System (FMS), coupler, SIS2 and other components, so using the source code from here will *not* necessarily work. [This repository](https://github.com/ashjbarnes/setup-mom6-nci) contains scripts for compiling MOM6, and includes a folder `src` containing submodules linked to the correct versions known to work with this package and its demos.
+The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. Note that GFDL's `MOM6_examples` repository uses older versions of the Flexible Modelling System (FMS), coupler, SIS2 and other components. [This repository](https://github.com/ashjbarnes/setup-mom6-nci) contains scripts for compiling MOM6, and includes a folder `src` containing submodules linked to the commits known to work with this package and its demos.
 
 ## Getting started
 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,9 @@ pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de05
 ## MOM6 Configuration and Version Requirements
 The package and demos were written for a coupled SIS2-MOM6 configuration, and may also work for an ocean only MOM6 run but this has not been tested. 
 
-The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. Note that GFDL's `MOM6_examples` repository uses older versions of the Flexible Modelling System (FMS), coupler, SIS2 and other components. [This repository](https://github.com/ashjbarnes/setup-mom6-nci) contains scripts for compiling MOM6, and includes a folder `src` containing submodules linked to the commits known to work with this package and its demos.
+The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. A forked version of the [`setup-mom6-nci`](https://github.com/ashjbarnes/setup-mom6-nci) repository contains scripts for compiling MOM6 and, furthermore, its [`src`](https://github.com/ashjbarnes/setup-mom6-nci/tree/setup-mom6/src) directory lists the particular commits that were used to compile MOM6 and its submodules for this package.
+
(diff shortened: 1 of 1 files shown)


ashjbarnes:
Just thought I'd get your opinion as to whether this approach makes sense. The package itself doesn't care about the version used (afaik) but the demo notebooks do

README.md
```diff
@@ -57,7 +57,10 @@ or
 ```bash
 pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de0566df329c4ea472002f7e
 ```
+## MOM6 Configuration and Version Requirements
+While the underlying package should work for coupled or ocean only configurations, the premade files for the demos are written for a coupled MOM6-SIS2 configuration.
 
+The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. Note that GFDL's `MOM6_examples` repository uses older versions of the Flexible Modelling System (FMS), coupler, SIS2 and other components, so using the source code from here will *not* necessarily work. [This repository]( https://github.com/ashjbarnes/setup-mom6-nci) contains scripts for compiling MOM6, and includes a folder `src` containing submodules linked to the correct versions known to work with this package and its demos.
```
navidcy, line 63:
```suggestion
The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. Note that GFDL's `MOM6_examples` repository uses older versions of the Flexible Modelling System (FMS), coupler, SIS2 and other components, so using the source code from here will *not* necessarily work. [This repository](https://github.com/ashjbarnes/setup-mom6-nci) contains scripts for compiling MOM6, and includes a folder `src` containing submodules linked to the correct versions known to work with this package and its demos.
```

I'm wondering: instead of writing comments like this that have the danger of getting outdated can't we just enforce dependency requirements and then everything is clear?

navidcy:
OK, fair point!

Then let's be a bit more specific.

- Let's avoid "as of January 2024". The commit come with a timestamp.
- Let's not say which versions won't work, e.g., "the source code from here will *not* necessarily work."
- Let's just say which versions work. E.g., "`regional-mom6 requires you to build mom6 executable from commit XXX"

navidcy (changes requested):
OK, fair point!

Then let's be a bit more specific.

- Let's avoid "as of January 2024". These statements become outdated... The commit anyway comes with a timestamp so everyone knows when the latest README was written.
- Let's not say which versions won't work, e.g., "the source code from here will not necessarily work."
- Let's just say which versions work. E.g., "`regional-mom6 requires you to build mom6 executable from commit XXX"

navidcy:
Also regarding:

> While the underlying package should work for coupled or ocean only configurations, the premade files for the demos are written for a coupled MOM6-SIS2 configuration.

Do we know for sure that it should work for coupled or ocean only configurations? Have we ever tested it? Let's not hypothesise if not. It's cleaner if we say

> All demos and tests were conducted for a coupled MOM6-SIS2 configuration. In principle, `regional-mom6` could also work for coupled or ocean-only MOM6 configurations, but we have not tested it.

Also, what is the difference of "coupled" and "coupled MOM6-SIS2"?

ashjbarnes:
>     * Let's just say which versions work. E.g., "`regional-mom6 requires you to build mom6 executable from commit XXX"

I'd rather not do this as it'll clutter up the readme a lot given there are many required packages. The link to my repo and the `src` directory there should be sufficient

ashjbarnes:
>     * Let's avoid "as of January 2024". These statements become outdated... The commit anyway comes with a timestamp so everyone knows when the latest README was written.

Yes but if we update another part of the readme then the timestamp will change and readers shouldn't then assume that the 'latest' source code refers to that timestamp. I think this way is clearest for now, especially with the link to the `src` directory of commits.

navidcy:
> > ```
> > * Let's just say which versions work. E.g., "`regional-mom6 requires you to build mom6 executable from commit XXX"
> > ```
> 
> I'd rather not do this as it'll clutter up the readme a lot given there are many required packages. The link to my repo and the `src` directory there should be sufficient

I only meant to write explicitly the versions of dependencies that cannot be properly added as package dependencies.

README.md
```diff
@@ -57,7 +57,10 @@ or
 ```bash
 pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de0566df329c4ea472002f7e
 ```
+## MOM6 Configuration and Version Requirements
+The package and demos were written for a coupled SIS2-MOM6 configuration, and may also work for an ocean only MOM6 run but this has not been tested. 
 
+The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. Note that GFDL's `MOM6_examples` repository uses older versions of the Flexible Modelling System (FMS), coupler, SIS2 and other components. [This repository](https://github.com/ashjbarnes/setup-mom6-nci) contains scripts for compiling MOM6, and includes a folder `src` containing submodules linked to the commits known to work with this package and its demos.
```
navidcy, line 63:
```suggestion
The current release of this package assumes the latest source code of all components needed to run MOM6 as of January 2024. The [`setup-mom6-nci`](https://github.com/ashjbarnes/setup-mom6-nci) repository contains scripts for compiling MOM6. The directory [`src`](https://github.com/ashjbarnes/setup-mom6-nci/tree/setup-mom6/src) of the above-mentioned repository lists the particular commits that were used to compile MOM6 and its submodules for this package.

Note that the commits used for MOM6 submodules (e.g., Flexible Modelling System (FMS), coupler, SIS2) are _not_ necessarily those used by the GFDL's [`MOM6_examples`](https://github.com/NOAA-GFDL/MOM6-examples) repository.
```

navidcy (approved):
OK, I only now realized that your repo listed all the commits used to compile MOM6 and its subcomponents.

I rephrased a bit. I think it's good now! It provides the info needed!

PR #95, opened 2024-02-16
Update installation instructions

navidcy:
Some rewording. Hopefully it's an enhancement.


navidcy:
@ashjbarnes the readme mentions:

> It's recommended that you get started with the example notebooks. In this case you'll want to clone this
entire repo so that you have access to the `demo` folder. This contains some template configuration files
for MOM6, and example notebooks that walk you through the package. 
> 
> ```bash 
> git clone git@github.com:COSIMA/regional-mom6.git
> pip install .
> ```

Is this the most updated latest recommendation? Why do people need to clone the repository? Just to get the notebook? They can simply download just the notebook, right?

navidcy:
For example, users can easily download the notebook from

https://regional-mom6.readthedocs.io/en/latest/demo_notebooks/reanalysis-forced.html

There is a button at the top right.

ashjbarnes:
The reason for the confusion is partly that some of the functionality requires the `premade run directory` folder. If you just `pip install` the package, you don't actually get these files. This is maybe because they sit in the main folder and not with the python code? I'm unsure. Any idea @angus-g ?

Anyway, my workaround has been that the `.setup_run_dir` method takes an argument to point to the location of the `regional-mom6` folder on disk, so that it can find these premade run directories. It's a bit clunky! And it was hard to explain in my updates to the install instructions

navidcy:
> The reason for the confusion is partly that some of the functionality requires the `premade run directory` folder. If you just `pip install` the package, you don't actually get these files. This is maybe because they sit in the main folder and not with the python code? I'm unsure. Any idea @angus-g ?

OK, I see. Is this premade run directory big in size or just few text files etc?

> Anyway, my workaround has been that the `.setup_run_dir` method takes an argument to point to the location of the `regional-mom6` folder on disk, so that it can find these premade run directories. It's a bit clunky! And it was hard to explain in my updates to the install instructions

Hm... well if that's what needed we need to explain it. Or give a step-by-step instructions.

ashjbarnes:
No the 

> > The reason for the confusion is partly that some of the functionality requires the `premade run directory` folder. If you just `pip install` the package, you don't actually get these files. This is maybe because they sit in the main folder and not with the python code? I'm unsure. Any idea @angus-g ?
> 
> OK, I see. Is this premade run directory big in size or just few text files etc?
> 
> > Anyway, my workaround has been that the `.setup_run_dir` method takes an argument to point to the location of the `regional-mom6` folder on disk, so that it can find these premade run directories. It's a bit clunky! And it was hard to explain in my updates to the install instructions
> 
> Hm... well if that's what needed we need to explain it. Or give a step-by-step instructions.

No it's a really small size on disk. I think ideally the `pip install` of the package should also pull down these premade run directories to keep things simple but I'm not sure how to do this

navidcy:
> No it's a really small size on disk. I think ideally the `pip install` of the package should also pull down these premade run directories to keep things simple but I'm not sure how to do this

But where would `pip install` save that directory? Wherever it saves the source code of the package? Or does this directory needs to be wherever the user wants to run the experiment?

ashjbarnes:
> > No it's a really small size on disk. I think ideally the `pip install` of the package should also pull down these premade run directories to keep things simple but I'm not sure how to do this
> 
> But where would `pip install` save that directory? Wherever it saves the source code of the package? Or does this directory needs to be wherever the user wants to run the experiment?

An old implementation of this was to use a relative path from the regional-mom6.py file. This was something like Path(__file__).parent / "demos". That works when you git clone the repository and then import from there, but doesn't work if you pip install (currently) because the demos folder isn't included. However, if we could guarantee that the files were included in the pip install, we could hardcode the correct relative path again in the `setup_rundir` function

navidcy:
As a user, do I need to have this `premade_run_dir` where I would run my experiment?

ashjbarnes:
> As a user, do I need to have this `premade_run_dir` where I would run my experiment?

no. the `setup_run_dir` just needs to know where to find the `premade_run_dirs` folder, which currently requires it to be git cloned because pip doesn't download it

angus-g:
See https://github.com/COSIMA/regional-mom6/tree/angus-g/package-data-install for how to include the demos in the package wheel.

navidcy:
So @angus-g, if we do what https://github.com/COSIMA/regional-mom6/commit/f064842b99b796c5282b24a7b08a41730cc3c6b6 suggests then this issue with the "premade_run_dir" that @ashjbarnes mentions above is taken care of?

angus-g:
> So @angus-g, if we do what [f064842](https://github.com/COSIMA/regional-mom6/commit/f064842b99b796c5282b24a7b08a41730cc3c6b6) suggests then this issue with the "premade_run_dir" that @ashjbarnes mentions above is taken care of?

Right, then from within the module code, `Path(__file__) / "premade_run_dir"` will exist and can be used as desired.

navidcy:
That sounds good!

OK, I did it. @ashjbarnes try this whenever and if it works fine then we can simplify the installation instructions!

ashjbarnes:
As it was, installing via `pip install .` in the module directory ran into a file not found error when trying to copy the premade run directories. I've used the `importlib.resources` module to return the absolute path to where the package is installed on the computer. This is used for finding the premade run directories

It also checks for the existence of this folder. If it can't find it, then it prompts the user to install via a clone of the repo so that the demo material is present

ashjbarnes (approved):
Tested and working with my last commits. I like the changes to documentation too!!

navidcy:
OK, can we simplify the installation instructions now? No need to ask people they need to clone the repo, right?

navidcy:
(I can do that, just wanted to double check that indeed users don't need to clone the repo...)

ashjbarnes:
Yes now the users should be able to install the repo like normal and still have access to the premade run directories so instructions can be updated

ashjbarnes:
Yeah so just install from github with pip install. I just need to fix up the notebooks to work with this new setup since they expect a hardcoded path to where the premade run dirs are

ashjbarnes:
Strangely installing from source and running this test locally works for me. Need to figure out what's going wrong in the test notebook environment!

navidcy:
@angus-g any insight here?

regional_mom6/regional_mom6.py
```diff
@@ -1474,8 +1474,11 @@ def setup_run_directory(
             overwrite_run_dir = premade_rundir_path / f"{surface_forcing}_surface"
             print(overwrite_run_dir)
             if not overwrite_run_dir.exists():
+                available = os.listdir(premade_rundir_path)
+                # Get all of the available surface forcing types
+                available = [x for x in available if os.path.isdir(premade_rundir_path / x)]
                 raise ValueError(
-                    f"Surface forcing {surface_forcing} not available. Please choose from {str(os.listdir(base_run_dir.parent))}."  ##Here print all available run directories
```
navidcy, line 1479: I was gonna say that this seemed odd...

ashjbarnes: Yeah just fixed that. Shouldn't change anything about the functionality but I figure improving the error messaging might help us out

angus-g:
Only thing I can think is that `importlib.resources` is picking up `regional_mom6` from the *current directory*, i.e. the one inside the repository which doesn't have the demos subdir...

ashjbarnes:
sorry for all the commits... I haven't found a good way to test the workflow locally yet. I've tried to do what @angus-g suggested but it just says that there's 'no such module'.

ashjbarnes:
Ok so the solution was to explicitly include all of the subdirectories of premade_run_directories. I'm still perplexed why doing a fresh install on a new conda env on gadi worked, but not within the docker container on github actions. Anyway I guess this is more robust!

ashjbarnes:
I've updated the install instructions so they no longer imply the need to git clone and then install to access the premade rundirs

navidcy:
Ok. Great!

navidcy:
It’d be nice if @angus-g’s pair of eyes had a look at this before we merge.

.github/workflows/testing.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           ln -s /data demos/PATH_TO_GLORYS_DATA
           ln -s /data demos/PATH_TO_GEBCO_FILE
           ln -s /build/FRE-NCtools/tools demos/PATH_TO_FRE_TOOLS
-          ln -s ../ demos/PATH_TO_REGIONAL_MOM6_CODE
+          # ln -s ../ demos/PATH_TO_REGIONAL_MOM6_CODE
```
navidcy, line 56: Why comment this. Either we need it or delete?

ashjbarnes: yeah can be deleted
```suggestion
```

README.md
```diff
@@ -66,18 +57,36 @@ or
 ```bash
 pip install git+https://github.com/COSIMA/regional-mom6.git@061b0ef80c7cbc04de0566df329c4ea472002f7e
 ```
+
+It's recommended that you get started with the example notebooks. In this case you'll want to clone this
```
navidcy, line 61: We need to update this to avoid suggesting cloning the repo

ashjbarnes: Yes I was wondering about that. Now that there's a better way to download the notebooks we can remove this comment too?
```suggestion

```

navidcy: Yeap. Do you want to fix this? You can mention that they can download the notebook from the docs

regional_mom6/regional_mom6.py
```diff
@@ -1459,24 +1459,27 @@ def setup_run_directory(
...
-        if surface_forcing != False:
-            overwrite_run_dir = (
-                Path(regional_mom6_path)
-                / "demos"
-                / "premade_run_directories"
-                / f"{surface_forcing}_surface"
+
+        # Define the locations of the directories we'll copy files across from. Base contains most of the files, and overwrite replaces files in the base directory.
+        base_run_dir = premade_rundir_path / "common_files"
+        if not os.path.exists(premade_rundir_path):
```
angus-g, line 1469:
```suggestion
        if not premade_rundir_path.exists():
```

regional_mom6/regional_mom6.py
```diff
@@ -1459,24 +1459,27 @@ def setup_run_directory(
...
             )
-            print(overwrite_run_dir)
+        if surface_forcing != False:
+            overwrite_run_dir = premade_rundir_path / f"{surface_forcing}_surface"
             if not overwrite_run_dir.exists():
+                available = os.listdir(premade_rundir_path)
+                # Get all of the available surface forcing types
+                available = [
+                    x for x in available if os.path.isdir(premade_rundir_path / x)
+                ]
```
angus-g, line 1480:
```suggestion
                available = [
                    x for x in premade_rundir_path.iterdir() if x.is_dir()
                ]
```

regional_mom6/regional_mom6.py
```diff
@@ -1459,24 +1459,27 @@ def setup_run_directory(
...
-                / f"{surface_forcing}_surface"
+
+        # Define the locations of the directories we'll copy files across from. Base contains most of the files, and overwrite replaces files in the base directory.
+        base_run_dir = premade_rundir_path / "common_files"
+        if not os.path.exists(premade_rundir_path):
+            raise ValueError(
+                f"Cannot find the premade run directory files at \n{premade_rundir_path}\n. Something is not right about how the package has been installed as these files are missing!"
             )
-            print(overwrite_run_dir)
+        if surface_forcing != False:
```
angus-g, line 1473: Why not:

```suggestion
        if surface_forcing:
```

angus-g (commented):
Looks pretty good. I guess there's still the outstanding change to avoid recommending a clone. Otherwise, just a few places where `os.path` should be replaced by pathlib directly.

regional_mom6/regional_mom6.py
navidcy: Yeah, @ashjbarnes. Writing `if surface_forcing != False` will execute if `surface_forcing = nothing` or `surface_forcing = "silly_forcing"`. Was this intentional?

navidcy: OK, @angus-g I see that how this method is called `surface_forcing` could be indeed a string `"era5"`.
So what @ashjbarnes did here is intentional.

It is prone for mistakes though since, at least myself, when I see a variable that can be `False` I assume it's a boolean so `True` or `False` are the only allowed values. I'll open an issue and we can discuss there and fix in a different  PR.

navidcy: see https://github.com/COSIMA/regional-mom6/issues/134

angus-g: I think in that case it should be a string value or `None`.

PR #97, opened 2024-02-16
Reorder doc contents

navidcy:
Bring examples up


ashjbarnes (approved):
I agree- floating the examples up showcases what the package does more easily

PR #98, opened 2024-02-16
Fix `_make_hgrid` docstring ambiguity + Enhance format + clarity of various docstrings

### LINKED ISSUE #52: Docstring regarding the resolution in latitudinal directions is slightly ambiguous

navidcy:
This PR enhances the formatting (e.g., using double backticks to enclose variable names) and some phrasing of the docstrings.

Also, closes #52.

I still have some questions/confusion regarding the docstring so I opened this PR to discuss and deal with them.


navidcy:
Outstanding questions:

- the grid as is generated has constant Δλ and Δφ = Δλ*cos(mean(φ)), right?
- what do we mean by "This is very simple but suitable for small domains."? what is "small" domain? let's be specific otherwise it's vague. For example, I have no idea at this point.

regional_mom6/regional_mom6.py
```diff
@@ -569,32 +572,33 @@ def rectangular_hgrid(λ, φ):
...
-        depth (float): Depth of the domain.
-        mom_run_dir (str): Path of the MOM6 control directory.
-        mom_input_dir (str): Path of the MOM6 input directory, to receive the forcing files.
-        toolpath (str): Path of FREtools binaries.
-        gridtype (Optional[str]): Type of horizontal grid to generate, only ``even_spacing`` is currently supported.
-        ryf (Optional[bool]): Whether the experiment runs 'repeat year forcing'. Otherwise assumes inter-annual forcing.
-        read_existing_grids (Optional[Bool]): Instead of generating them again, reads the grids and ocean mask from the inputdir and rundir. Useful for modifying or troubleshooting experiments.
+        `xextent` (Tuple[float]): Extent of the region in longitude in degrees.
+        `yextent` (Tuple[float]): Extent of the region in latitude in degrees.
+        `daterange` (Tuple[str]): Start and end dates of the boundary forcing window.
```
navidcy, line 591: @ashjbarnes, let's add an example in the docstring. Is what I suggest below OK?

```python
        `daterange` (Tuple[str]): Start and end dates of the boundary forcing window. For example: `('2002-11-01', '2002-11-30')`
```

regional_mom6/regional_mom6.py
```diff
@@ -483,27 +483,30 @@ def quadrilateral_areas(lat, lon, R=1):
...
+        xarray.Dataset: An FMS-compatible `hgrid` that includes all required attributes.
     """
 
-    R = 6371e3  # mean radius of the Earth
+    R = 6371e3  # mean radius of the Earth; https://en.wikipedia.org/wiki/Earth_radius
 
-    dλ = λ[1] - λ[0]  # assuming that longitude is uniformly spaced
+    # compute longitude spacing and ensure that longitudes are uniformly spaced
+    dλ = λ[1] - λ[0]
+    assert np.max(np.diff(λ) - dλ) < 1e-14
```
navidcy, line 509: @ashjbarnes I added this `assert` here so that we don't just rely on "trusting" that the users would have read the docstring and provided with input as we expect. If there are other places in the code that something is assumed then we can ensure that it is the case with an `assert` statement so that it errors if the assumption is not as we expect.

regional_mom6/regional_mom6.py
```diff
@@ -270,25 +274,30 @@ def motu_requests(
...
+            ``"GLOBAL_MULTIYEAR_PHY_001_030-TDS"``.
+        productid (Optional[str]): Data product within the chosen service. Default:
+            ``"cmems_mod_glo_phy_my_0.083_P1D-m"``.
+
+    Return:
         str: A bash script which will call ``motuclient`` to invoke the data requests.
-
     """
 
     if type(segs) == str:
```
navidcy, line 303: is this better way for doing this @angus-g ?

```suggestion
    if isistance(sets, str)
```

angus-g: Yeah, it is.

regional_mom6/regional_mom6.py
```diff
@@ -69,7 +72,7 @@ def ap2ep(uc, vc):
         uc: complex tidal u velocity
         vc: complex tidal v velocity
 
-    Returns:
+    Return:
```
angus-g, line 75: I think this should stay as it was (it's in the verb sense, not the noun sense), and same for the rest of the file.

navidcy: True! I just notice this morning.

Interestingly, the `make html` for the docs replaces this with "Returns" anyway! But I'll change it in the code, no need to have "make html" fix our mistakes!

navidcy: done by a311855

regional_mom6/regional_mom6.py
ashjbarnes: Yeah I like that! Original was a bit clunky

ashjbarnes (approved):
Looks great! Thanks Navid

navidcy:
> Thanks @ashjbarnes.
> 
> OK, there is only one outstanding question: What do we mean by
> 
> > This is very simple but suitable for small domains.
> 
> in the `_make_hgrid` docstring. If we clarify that then we can also close #52. Perhaps just a phrase is needed, e.g.,
> 
> > This is very simple but suitable for small domains, e.g., for domains which the latitude range is small enough so that `cos(φ_North_Side)` is not much different from `cos(φ_South_Side)`.
> 
> But I'm not sure that this is what we mean there to be honest; it's just a guess.

OK, I think I understood it --> [ae31888](https://github.com/COSIMA/regional-mom6/pull/98/commits/ae318886d72dc3629bbbd3b32872a57cda12091a)

navidcy:
So something happened and the test fail at the regridder? Did I break anything?

ashjbarnes:
It's crashing because 'lon' is no longer in the dataset. Perhaps it's because of some of the dimension renaming from https://github.com/COSIMA/regional-mom6/pull/98/commits/ae318886d72dc3629bbbd3b32872a57cda12091a ?

navidcy:
> It's crashing because 'lon' is no longer in the dataset. Perhaps it's because of some of the dimension renaming from [ae31888](https://github.com/COSIMA/regional-mom6/commit/ae318886d72dc3629bbbd3b32872a57cda12091a) ?

1. I had reverted the `nlayers` to `vlayers` in [29253d7](https://github.com/COSIMA/regional-mom6/pull/98/commits/29253d7fa5df9ec33fa273bd722589a4dba69621)

2. I reverted back everything and [still test fail](https://github.com/COSIMA/regional-mom6/actions/runs/7984349545/job/21801046518?pr=98#step:7:149)...

3. Locally tests pass.

Hm....

angus-g:
I think it's because xESMF pushed 0.8.3 to fix the xarray incompatibility, but it brought in [this change](https://github.com/pangeo-data/xESMF/compare/v0.8.2...v0.8.3#diff-e2be4696e3d09637f3876c3ff9b131bb02c6bb9b827c051125aeb36f73250341R1032-R1035) which for whatever reason doesn't work on that particular dataset.

navidcy:
omg... :) 
do we need a compat entry for xESMF now?

ashjbarnes:
wow.. ok good find Angus! I don't think the test is doing anything crazy. Perhaps the xesfm people have overlooked something..

navidcy:
OK, with the [compat entry](https://github.com/COSIMA/regional-mom6/pull/98/commits/781bbae7cf28580ca48c6cfcbc2643472c0455f8) test pass. Shall I merge this with the compat entry until xESMF and xarray sort it out @angus-g?

angus-g:
> Shall I merge this with the compat entry until xESMF and xarray sort it out

Sure, I guess. Nothing has been raised with xESMF yet, I'm just trying to put together a bug report for them.

Issue #96, opened 2024-02-16
Enhance `reanalysis-forced` notebook


navidcy:
Let's enhance the notebook `reanalysis-forced` at https://regional-mom6.readthedocs.io/en/latest/demo_notebooks/reanalysis-forced.html with some plots of, e.g, the domain or the initial condition etc.

Also let's add a thumbnail to appear at https://regional-mom6.readthedocs.io/en/latest/demos.html

angus-g:
Maybe moving discussion from #99 here: I've implemented a way to run the notebooks in the container on GitHub (where the model, tools, and input data are available), and provide that notebook to RTD. Example build is https://regional-mom6.readthedocs.io/en/angus-g-render-notebooks/demo_notebooks/reanalysis-forced.html, from the angus-g/render-notebooks branch.

At the moment, there's both the GitHub push webhook, and the RTDS action webhook firing for the same version, so it's not perfectly integrated until we completely switch over. There might be a little bit more to do in generating previews for PRs (for this demo, I had to manually activate the branch on RTD).

navidcy:
This looks great!

aekiss:
> Also let's add a thumbnail to appear at https://regional-mom6.readthedocs.io/en/latest/demos.html

thumbnail still missing

navidcy:
Yes. Related to this issue is https://github.com/COSIMA/regional-mom6/issues/119 which if resolved then we could have the notebook cells evaluated locally and push.

PR #99, opened 2024-02-16
Some formatting beautifications in demos

navidcy:


navidcy:
I don't understand why tests fail with the changes I did. Any insight?

ashjbarnes:
Looks like the deletion of a curly brace:

image

navidcy:
Nice catch!!!!

navidcy:
Are the notebooks cell's executed by the docs?

If so, then why there is no plot showing up e.g. at https://regional-mom6--99.org.readthedocs.build/en/99/demo_notebooks/reanalysis-forced.html#Step-4:-Set-up-bathymetry

?

navidcy:
The notebook cells are executed by the tests but not by the docs?

cc @angus-g

ashjbarnes:
> The notebook cells are executed by the tests but not by the docs?
> 
> cc @angus-g

The tests execute all the cells but skip the plotting since matplotlib isn't included in the package. We discussed adding it but decided not to since it's not actually used by the package, just the notebooks.

If the docs were able to execute the notebooks (I'm not sure they can) then they'd also not have matplotlib loaded if just using the python environment required by the package

navidcy:
OK, not sure how things work in Python but if we can have the Docs execute the cells + include matplotlib in the docs environment that would be great!

We definitely don't want to have matplotlib as a dependency of the package. But I don't see why not we can't have it in the docs environment. We could probably use it elsewhere, e.g., generate plots in a future tutorial that we might wanna have in the docs...

ashjbarnes:
Yeah agreed. Not sure how to do this. Perhaps the docker image could be modified to come with python3 inc matplotlib, and then our package is installed on top of all this? But then I'm not sure how to use the docker image to build the docs. 

Another option is just to re-commit the notebooks with the plotting cells executed? There's probably no need to have them run by the docs as well as by the tests right?

navidcy:
Having them run by the docs ensures that the output is up to date with the latest package version!

angus-g:
Why not just install matplotlib in the testing workflow? Like:

https://github.com/COSIMA/regional-mom6/blob/71e30862cf71a9d7bfaa3a93af0aff4b1b5245aa/.github/workflows/testing.yml#L38-L43

ashjbarnes:
> Having them run by the docs ensures that the output is up to date with the latest package version!

Ok but running the demos requires the dummy input data, so involves running them on the docker image. There are also a couple of extra steps setting up symlinks too that get the notebook to work with the "dummy" directories

https://github.com/COSIMA/regional-mom6/blob/71e30862cf71a9d7bfaa3a93af0aff4b1b5245aa/.github/workflows/testing.yml#L53C1-L57C129

navidcy:
> > Having them run by the docs ensures that the output is up to date with the latest package version!
> 
> Ok but running the demos requires the dummy input data, so involves running them on the docker image. There are also a couple of extra steps setting up symlinks too that get the notebook to work with the "dummy" directories
> 
> https://github.com/COSIMA/regional-mom6/blob/71e30862cf71a9d7bfaa3a93af0aff4b1b5245aa/.github/workflows/testing.yml#L53C1-L57C129

OK, if it's way cumbersome then we can just push the notebooks with the cells executed for now and see how we go about it later?

navidcy:
> Ah right, I understand. I do have some ways to do this, but don't have time until tomorrow during normal work hours.

No rush at all!

demos/access_om2-forced.ipynb
navidcy, line 1: > **NOTE: this is hardcoded for the year of 1990, which corresponds to output files `1077`-`1082`. If you want to modify, you'll need to choose the right path to the year of your choice, or use the COSIMA cookbook to locate your data files**

Why don't we use the cookbook here?

demos/access_om2-forced.ipynb
```diff
@@ -6,7 +6,7 @@
    "source": [
```
navidcy, line 6: <pre class="ql-syntax" spellcheck="false">Line #6.            yt_ocean = slice(yextent[0] - 0.2, yextent[1] + 0.2)).isel(time = slice(0, 5))
</pre><p> </p><p>this reads like 4 months to me? <code>slice(0, 5)</code>  or is this 3 months??</p><p>if it's not 3 months let's correct what the paragraph above claims?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/COSIMA/regional-mom6/pull/99/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

demos/reanalysis-forced.ipynb
```diff
@@ -34,7 +34,7 @@
    "outputs": [],
```
navidcy, line 34: <pre class="ql-syntax" spellcheck="false">Line #7.    daterange = ["2003-01-01 00:00:00", "2003-01-05 00:00:00"] ## 2003 is a good compromise for GLORYs and JRA forcing as they overlap. JRA ends in 2012, GLORYS starts in 1993
</pre><p> </p><p>JRA ends in Jan. 2024 so I don't understand this comment.</p><p> </p><p>Anyway the comment can be sharpened or even deleted? Because when the user comes to this point they may be wondering what does JRA has to do with this here? Are we talking about various reanalyses products? Then what about ERA5... etc? (My point is that the comment probably creates more questions than it answers.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/COSIMA/regional-mom6/pull/99/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

navidcy:
> Ah right, I understand. I do have some ways to do this, but don't have time until tomorrow during normal work hours.

This is related to #96 btw.

ashjbarnes (approved):
Oops I thought I already approved this one! We'll need to merge with #122

demos/access_om2-forced.ipynb
navidcy: "crashed cryptically"! :)

demos/reanalysis-forced.ipynb
```diff
@@ -368,7 +364,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.13"
+   "version": "3.10.6"
```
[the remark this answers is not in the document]
navidcy: not sure, I didn't change this manually... 😳

demos/access_om2-forced.ipynb
navidcy, line 6: <p>I see that <code>make_solo_mosaic</code> congratulates us with a typo ;) ("congradulation")</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/COSIMA/regional-mom6/pull/99/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

navidcy:
OK, I'll just update the ACCESS-OM2 notebook to align with the latest API changes we did and don't bother much with it.

navidcy:
OK, I merged this.

We'd still like to be able to have the docs run the notebook live as the docs are generated (see https://github.com/COSIMA/regional-mom6/issues/96) to ensure that the output is up to date with the latest version of the package.

PR #104, opened 2024-02-16
Remove unused imports

### LINKED ISSUE #103: Some imports are never used?
We import

https://github.com/COSIMA/regional-mom6/blob/3e9844286db015e17940bbbacc2c35f7ddc82e23/regional_mom6/regional_mom6.py#L2

https://github.com/COSIMA/regional-mom6/blob/3e9844286db015e17940bbbacc2c35f7ddc82e23/regional_mom6/regional_mom6.py#L5

https://github.com/COSIMA/regional-mom6/blob/3e9844286db015e17940bbbacc2c35f7ddc82e23/regional_mom6/regional_mom6.py#L12

but I don't see where we use them.

Are they actually needed?

navidcy:
I think some of the imports are not used so I remove them here and see what happens.

Closes #103

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1,7 +1,6 @@
 import numpy as np
 from pathlib import Path
 import dask.array as da
-import numpy as np
 import xarray as xr
 import xesmf as xe
 import subprocess
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1,21 +1,19 @@
 import numpy as np
-from itertools import cycle
 from pathlib import Path
 import dask.array as da
-import dask.bag as db
 import numpy as np
 import xarray as xr
 import xesmf as xe
 import subprocess
 from scipy.ndimage import binary_fill_holes
 import netCDF4
-from dask.distributed import Client, worker_client
 from dask.diagnostics import ProgressBar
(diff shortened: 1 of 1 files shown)


navidcy:
hm... test pass so they were probably no needed?

regional_mom6/regional_mom6.py
```diff
@@ -1,21 +1,19 @@
 import numpy as np
-from itertools import cycle
 from pathlib import Path
 import dask.array as da
-import dask.bag as db
 import numpy as np
```
angus-g, line 4: We also import numpy twice!

angus-g (approved):
Yep, good catch! I've just dropped out the duplicated numpy import too.

regional_mom6/regional_mom6.py
```diff
@@ -1,21 +1,18 @@
...
-from itertools import cycle
```
navidcy, line 2: @angus-g, is `itertools` an unnecessary dependency of the package?

navidcy: hm... I guess `itertools` is part of Python so we are all good!

angus-g: No, it's a Python built-in.

navidcy: I'm learning... :)

PR #108, opened 2024-02-19
Add `xarray` upper bound to avoid v2024.02.0

navidcy:
After the [latest xarray release](https://github.com/pydata/xarray/releases/tag/v2024.02.0) errors started appearing in the CI. See, for example, https://github.com/COSIMA/regional-mom6/actions/runs/7957078139/job/21719091616#step:7:226

See https://github.com/pydata/xarray/issues/8769

Adding an upper bound in `pyproject` to preclude the xarray v2024.02.0 resolves the issue for now until a patch release comes out from xarray.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
   "netCDF4",
   "numpy>=1.17.0",
   "scipy>=1.2.0",
-  "xarray",
+  "xarray<2024.02.0",
   "xesmf>=0.8",
   "f90nml>=1.4.1",
 ]


angus-g:
The root of this is xESMF needing to upgrade to look for `is_dask_collection` in `xarray.core.utils` not `xarray.core.pycompat` (https://github.com/pangeo-data/xESMF/issues/338)

navidcy:
oh I see! is there a better solution than adding a compat entry for xarray?

navidcy:
If you can comment on the issue I opened in xarray it might be helpful. Perhaps that needs to close then..

angus-g:
> is there a better solution than adding a compat entry for xarray?

No, I think this is still the way to go.

navidcy:
> > is there a better solution than adding a compat entry for xarray?
> 
> No, I think this is still the way to go.

OK, cool!

angus-g:
But I wonder if there's a way to remind us to unpin when there's a new xESMF release...

navidcy:
you hope that because we mentioned the xESMF issue we'll now get notifications when things update there and we'll see when it's resolved, right?

navidcy:
I actually subscribed to the issue to make sure.

angus-g (approved):
Yep, let's go with this for now.

PR #109, opened 2024-02-20
Adds some examples in docstrings with doctests

navidcy:


ashjbarnes:
When moving these functions to utils does it change anything on the user end? As in will they need to import extra things from utils rather than just `import regional-mom6 as rmom6`?

navidcy:
Good point. Let me look at this

navidcy:
These functions are not used by the users, right? As if, I haven't seen them used in any of the notebooks.

If the users wanna use them, then true, how you'd import them would change after this PR. E.g. instead of

```python
from regional_mom6 import angle_between
```

we'd have to do

```python
from regional_mom6.utils import angle_between
```

navidcy:
You can see in the Docs/API where to find them.

Btw the tests fail in this PR as well... :)

ashjbarnes:
Oh right what's breaking the tests? Have you traced back what's removed that "lon" coordinate from the bathymetry interpolation function?

navidcy:
I haven't been able to do so. Test pass locally on my machine...

navidcy:
OK, the xESMF compat entry resolved the issue

ashjbarnes (approved):
Looks good!

PR #111, opened 2024-02-24
Unpin xarray and xesmf

navidcy:
Try to remove the pinning of xarray and xesmf.


navidcy:
hm... is it expected that test on python 3.9 stall?

navidcy:
OK, they don't hang... it was something random probably.

angus-g (approved):
> OK, they don't hang... it was something random probably.

I think they do randomly hang, I've seen that a few times too... But I guess it's hard to say what's going wrong!

Anyway, let's get this in! Such a speedy response from upstream.

Issue #32, opened 2023-06-14
More verbose method/function naming + push for explicit keyword args vs positional args


navidcy:
I suggest use of more verbose naming. E.g.

```
def dz(npoints,ratio,target_depth,min_dz = 0.0001,tolerence = 1):
    blah blah ...
```

```
def vertical_grid_spacing(*, number_of_points, ratio, target_depth, minimum_vertical_grid_spacing = 1e-3, tolerance = 1):
    blah blah ...
```

navidcy:
Using kwargs with verbose names makes the user experience sort of seamless. Our demo/examples scripts don't have to include comments to explain what each thing is, e.g.,

```
# calculate the vertical gridspacing
dz_grid = dz(100, # number of points
             30, # the ratio
             5000, # the target depth
             min_dz = 0.0001,
             tolerence = 1)
```

becomes the self-explanatory

```python
dz = vertical_grid_spacing(number_of_points = 100,
                           ratio = 30,
                           target_depth = 5000,
                           minimum_vertical_grid_spacing = 1e-3,
                           tolerance = 1)
```

navidcy:
Some remarks on current names:


- https://github.com/COSIMA/mom6-regional/blob/e3e59a6193a684cfeeaf3b31eb13b80c0dea21bd/mom6_regional/mom6_regional.py#L31

  `nicer_..;` implies comparison with something else and creates question. It's a bit esoteric. We don't need to talk down to other slicers (which ever those may be). I'd say we name this `slicer` or even better something more descriptive of what this function does

- https://github.com/COSIMA/mom6-regional/blob/e3e59a6193a684cfeeaf3b31eb13b80c0dea21bd/mom6_regional/mom6_regional.py#L248

  `dz` implies `vertical_grid_spacing` so I'd suggest we rename this function like that. But from reading the docstring this function does not return the vertical grid spacing but rather a particular vertical level configuration with a hyperbolic tangent-like spacing configuration. Let's rename this to `tanh_spacing_vertical_levels` or something?

- https://github.com/COSIMA/mom6-regional/blob/e3e59a6193a684cfeeaf3b31eb13b80c0dea21bd/mom6_regional/mom6_regional.py#L321

  This reads better as `rectangular_horizontal_grid`. But it's not even rectangular so why do we describe it as "rectangular"?

ashjbarnes:
[...]

PR #112, opened 2024-02-27
Enhance (and rename) `nicer_slicer` -> `longitude_slicer`

navidcy:
Closes https://github.com/COSIMA/regional-mom6/issues/32


navidcy:
I noticed

https://github.com/COSIMA/regional-mom6/blob/544eb434ad1029231b436f7a7891f4ad0b92692a/regional_mom6/regional_mom6.py#L207

Let's add an `assert` to ensure our assumption is correct!

navidcy:
I also noticed 

https://github.com/COSIMA/regional-mom6/blob/544eb434ad1029231b436f7a7891f4ad0b92692a/regional_mom6/regional_mom6.py#L161

Why isn't `[0, 360]` not part of the "commonly used domains"?

navidcy:
I think `nicer_slicer` should be renamed to `longitude_slicer`, right?

navidcy:
Question:

Here:

https://github.com/COSIMA/regional-mom6/blob/544eb434ad1029231b436f7a7891f4ad0b92692a/regional_mom6/regional_mom6.py#L165-L176

we mention `oldx` but I don't see any `oldx` in the source code. Is the `oldx` only for reference in the docstring or was there an `oldx` variable at some point that got deprecated?

navidcy:
Ideally this PR should also deal with #118

ashjbarnes (approved):
Looks good. Might need to merge renaming to longitude_slicer will merge nicely with #122

PR #106, opened 2024-02-16
Clean up the vertical thickness profile generator (`dz_hyperbolictan`) + modify `_make_vgrid` to include `zi = 0`

navidcy:
This PR simplifies `dz_hyperbolictan`. The parameter `ratio` is now defined as `thickness bottom layer / thickness top layer`. Thus there is no need for negative values of `ratio` that seemed only a matter of "convention". Now `ratio > 1` implies that the bottom layer is thicker, `ratio < 1` that the top is thicker and `ratio = 1` implies equally-spaced layers.

The PR also adds examples in the docstring that demonstrate the function's behavior. Have a look at 

https://regional-mom6--106.org.readthedocs.build/en/106/api.html#regional_mom6.regional_mom6.hyperbolictan_thickness_profile

The PR also adds a bit more explanation regarding what "hyperbolic tangent" is in this context. Closes #53

Additionally, this PR simplifies the vertical grid generation and fixes an issue that the interface `zi = 0` was previously not included in the vertical grid. Closes #107.

Issue #21, opened 2023-05-31
key error in generating topo


AndyHoggANU:
I've made a start on testing out the demo script, but I get the following error when generating the topography:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File /g/data/hh5/public/apps/miniconda3/envs/analysis3-22.10/lib/python3.9/site-packages/xarray/core/dataset.py:1347, in Dataset._construct_dataarray(self, name)
   1346 try:
-> 1347     variable = self._variables[name]
   1348 except KeyError:

KeyError: 'elevation'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
Cell In[8], line 1
----> 1 expt.bathymetry(
      2     '/g/data/ik11/inputs/GEBCO_2022/GEBCO_2022.nc',
      3     {"xh":"lon",
      4      "yh":"lat",
      5      "elevation":"elevation"}, ## Again this dictionary just maps mom6 variable names to what they are in your topog.
      6     )

File /home/149/ab8992/cosima_regional/mom6-regional-scripts/regional_library.py:702, in experiment.bathymetry(self, bathy_path, varnames, fill_channels, minimum_layers, maketopog)
    698 topog = xr.open_dataset(self.mom_input_dir + "topog_raw.nc", engine='netcdf4')
    700 ## Ensure correct encoding
    701 topog = xr.Dataset(
--> 702     {"depth":(["ny","nx"],topog[varnames["elevation"]].values)}
    703 )
    704 topog.attrs["depth"] = "meters"
    705 topog.attrs["standard_name"] = "topographic depth at T-cell centers"

File /g/data/hh5/public/apps/miniconda3/envs/analysis3-22.10/lib/python3.9/site-packages/xarray/core/dataset.py:1438, in Dataset.__getitem__(self, key)
   1436     return self.isel(**key)
   1437 if utils.hashable(key):
-> 1438     return self._construct_dataarray(key)
   1439 if utils.iterable_of_hashable(key):
   1440     return self._copy_listed(key)

File /g/data/hh5/public/apps/miniconda3/envs/analysis3-22.10/lib/python3.9/site-packages/xarray/core/dataset.py:1349, in Dataset._construct_dataarray(self, name)
   1347     variable = self._variables[name]
   1348 except KeyError:
-> 1349     _, name, variable = _get_virtual_variable(self._variables, name, self.dims)
   1351 needed_dims = set(variable.dims)
[...]

PR #113, opened 2024-02-27
Revise `experiment` construction to utilize keyword arguments to enhance clarity

navidcy:
This PR changes the experiment constructor, more towards to what #32 is suggesting.

The PR introduces a breaking change in the API. 

### before this PR

```python
import regional_mom6 as rmom6

yextent = [-48, -38.95] ## latitude
xextent = [143, 150] ## longitude
daterange = ["2003-01-01 00:00:00", "2003-01-05 00:00:00"]
inputdir = Path(f"mom6_input_directories/{expt_name}/")
rundir = Path(f"mom6_run_directories/{expt_name}/")
toolpath = Path("PATH_TO_FRE_TOOLS")

expt = rmom6.experiment(
    xextent,
    yextent,
    daterange,
    0.05,  # Horizontal Resolution
    75,    # Number of vertical layers
    10,    # Ratio of largest to smallest vertical layer.
    4500,  # Depth of simulation
    rundir,
    inputdir,
    toolpath
)
```

### after this PR (preliminary, kwarg names to be discussed)

```python
import regional_mom6 as rmom6

expt = rmom6.experiment(
    longitude_extent = [-48, -38.95],
    latitude_extent = [143, 150],
    date_range = ["2003-01-01 00:00:00", "2003-01-05 00:00:00"],
    resolution = 0.05,
    number_vertical_layers = 75,
    layer_thickness_ratio = 10,
    depth = 4500,
    mom_run_dir = Path(f"mom6_input_directories/{expt_name}/"),
    mom_input_dir = Path(f"mom6_run_directories/{expt_name}/"),
    toolpath_dir = Path("PATH_TO_FRE_TOOLS")
)
```

The latter is really self-explanatory; no need to add comments next to every argument to remind people what each arg is.


navidcy:
Is there a reason that the private method here:

https://github.com/COSIMA/regional-mom6/blob/0143c6405e6253fede7fef467d4132cb6182215f/regional_mom6/regional_mom6.py#L593

takes `grid_type` as an argument and doesn't infer it from `self`?

navidcy:
Another question: 

The docstring here

https://github.com/COSIMA/regional-mom6/blob/0d6f617fa1b5c2fca61518a76933dd5922ca924c/regional_mom6/regional_mom6.py#L520

mentions that only "``xextent``, ``yextent``, ``daterange``, and ``resolution``" are required but I see more args that don't have default values so we need to update it, right?

angus-g:
I don't think we need to repeat the argument list in the description (which also specifies which arguments are optional), and at least `xextent` and `yextent` have been renamed since.

navidcy:
> I don't think we need to repeat the argument list in the description (which also specifies which arguments are optional), and at least `xextent` and `yextent` have been renamed since.

Agree, we can shorten the description!

demos/access_om2-forced.ipynb
```diff
@@ -422,28 +422,26 @@
    "source": [
```
angus-g, line 422: <pre class="ql-syntax" spellcheck="false">Line #9.    inpu_tdir = f"{scratch}/regional_mom6_configs/{expt_name}/"
</pre><p>oops?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/COSIMA/regional-mom6/pull/113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

angus-g, line 422: <pre class="ql-syntax" spellcheck="false">Line #21.    for i,orientation in enumerate(["south", "north", "west", "east"]):
</pre><p><code>for i, orientation ...</code> (i.e. add a space, as you did in the reanalysis-forced notebook)?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/COSIMA/regional-mom6/pull/113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

navidcy: nice catch!

navidcy: done

navidcy:
Do you like the change of `args` to `kwargs` @angus-g?

angus-g:
> Do you like the change of `args` to `kwargs` @angus-g?

I think so, given there are so many arguments (and 4 numbers in a row), it definitely increases clarity.

ashjbarnes:
> Is there a reason that the private method here:
> 
> https://github.com/COSIMA/regional-mom6/blob/0143c6405e6253fede7fef467d4132cb6182215f/regional_mom6/regional_mom6.py#L593
> 
> takes `grid_type` as an argument and doesn't infer it from `self`?

No there's no reason for this actually!

ashjbarnes (approved):
This looks good. Are the changes still API changing if the arguments remain in the same order now that they're kwargs? If so this should also trigger a new tag release alongside #106

navidcy:
Yeap, we should include this in a new release.
It's a user-facing API change since the users need to provide keywords now; their previous notebook won't work out of the box.

Issue #115, opened 2024-02-28
Typo or is it really "quick quick mosaic"?


navidcy:
Is this what the MOM6 terminology is or do we have a typo here?

https://github.com/COSIMA/regional-mom6/blob/b0e13c5f6a45eec9451ea2165f19ae0fe9176aed/regional_mom6/regional_mom6.py#L1404

ashjbarnes:
Oh yeah definitely typo!

PR #123, opened 2024-03-12
Tolerence of 10^-14 for even spacing too strict

ashjbarnes:
The assertion that the grid spacing must not vary by more than 10^-14 fails for high enough resolution runs. To replicate, set `resolution = 0.0125`, with an `xextent` and `yextent` of 30 degrees or so. What happens is that because the memory is symmetric, we need to have an odd number of longitude points so that it starts and finishes with a q point. I guess that dividing by an odd number results in some numerical error, which for high enough resolution exceeds 10^-14.

I suggest changing this tolerance to 10^-10

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -539,7 +539,7 @@ def rectangular_hgrid(λ, φ):
     dλ = λ[1] - λ[0]
 
     assert (
-        np.max(np.diff(λ) - dλ) < 1e-10
+        np.allclose(np.diff(λ), dλ * np.ones(np.size(λ)-1))
     ), "provided array of longitudes must be uniformly spaced"
 
     # dx = R * cos(φ) * np.deg2rad(dλ) / 2
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -206,8 +206,8 @@ def longitude_slicer(data, longitude_extent, longitude_coords, buffer=2):
 
         dλ = data[lon][1] - data[lon][0]
 
-        assert (
-            np.max(np.abs(np.diff(data[lon]) - dλ)) < 1e-14
+        assert np.allclose(
+            np.diff(data[lon]), dλ * np.ones(np.size(λ) - 1)
         ), "provided longitude coordinate must be uniformly spaced"
 
         assert np.isclose(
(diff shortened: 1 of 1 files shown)


navidcy:
For the case you suggested above I get:

```python
>>> λ = np.linspace(0, 30, 30*8+1)
>>> dλ = λ[1] - λ[0]
>>> np.max(np.diff(λ) - dλ) < 1e-14
True
```

So perhaps the culprit is something else?

navidcy:
> I guess that dividing by an odd number...

where is the division with the number of longitude points happen?

ashjbarnes:
> > I guess that dividing by an odd number...
> 
> where is the division with the number of longitude points happen?

```
import numpy as np

yextent = [-56,-26]
xextent = [142,180]
resolution = 0.0125

nx = int((xextent[1] - xextent[0]) / (resolution / 2))
print(nx)
if nx % 2 != 1:
    nx += 1
print(nx)
λ = np.linspace(
    xextent[0], xextent[1], nx
)  # longitudes in degrees
dλ = λ[1] - λ[0]

print(np.max(np.diff(λ) - dλ))
assert (
    np.max(np.diff(λ) - dλ) < 1e-14
), "provided array of longitudes must be uniformly spaced"
```
This test fails on numpy `1.21.5`, python `3.10.12`

regional_mom6/regional_mom6.py
```diff
@@ -440,7 +440,7 @@ def rectangular_hgrid(λ, φ):
     dλ = λ[1] - λ[0]
 
     assert (
-        np.max(np.diff(λ) - dλ) < 1e-14
+        np.max(np.diff(λ) - dλ) < 1e-10
```
navidcy, line 443: The function `numpy.allclose` takes into account numerical tolerances better.

```suggestion
        np.allclose(np.diff(λ), dλ * np.ones(np.size(λ)-1))
```

Indeed the snippet you posted was failing for me also. The code I suggest above works for even down to `resolution = 0.000001`...!

navidcy (changes requested):
I suggested an alternative that doesn't hardcode any tolerance value :)

navidcy:
We should also update

https://github.com/COSIMA/regional-mom6/blob/aa624c951df49cc8d95e7793ec4ffff98ba93b2f/regional_mom6/regional_mom6.py#L209-L211

as well.

navidcy:
perhaps we can resolve #115 here while we are at it?

navidcy (approved):
lgtm!

PR #127, opened 2024-03-13
Ensure assumptions are met in `rectangular_hgrid`

navidcy:
This PR asserts that `λ` and `φ` are monotonically increasing, as described in the docstring.

Note: I'm not sure where this assumption is used though... Is it used only when the `dλ` is calculated?


dhruvbhagtani (approved):
Looks good!

PR #133, opened 2024-03-27
Use verbosely-silly names when creating data arrays for tests

### LINKED ISSUE #114: `deepness`
Is `deepness` the property/variable name that MOM6 expects or is it an ad-hoc choice here?

https://github.com/COSIMA/regional-mom6/blob/b0e13c5f6a45eec9451ea2165f19ae0fe9176aed/tests/test_expt_class.py#L158

navidcy:
This PR uses really silly names and adds a remark to avoid any confusion created, e.g., see #114.

Closes #114


ashjbarnes (approved):
silly_approve

Issue #134, opened 2024-03-28
`setup_run_directory` includes arguments that are either boolean or strings


navidcy:
The `setup_run_directory` method

https://github.com/COSIMA/regional-mom6/blob/c2c6dda9077a3f8209118a562df1f1421870fa1d/regional_mom6/regional_mom6.py#L1438

uses `surface_forcing` argument that seems to take either `False` or a string, e.g., `"era5"`.

This is prone for mistakes though since, at least myself, when I see a variable that can be `False` I assume it's a boolean so `True` or `False` are the only allowed values.

_Originally posted by @navidcy in https://github.com/COSIMA/regional-mom6/pull/95#discussion_r1542377022_

navidcy:
This was closed by #135

PR #135, opened 2024-03-28
Change default arg from False to None in the setup run dir

ashjbarnes:
PR closes issue #134

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1438,7 +1438,7 @@ class experiment:
 
     def setup_run_directory(
         self,
-        surface_forcing=False,
+        surface_forcing=None,
         using_payu=False,
         overwrite=False,
     ):
@@ -1470,7 +1470,7 @@ class experiment:
             raise ValueError(
                 f"Cannot find the premade run directory files at \n{premade_rundir_path}\n. Something is not right about how the package has been installed as these files are missing!"
             )
-        if surface_forcing != False:
+        if type(surface_forcing) == str:
             overwrite_run_dir = premade_rundir_path / f"{surface_forcing}_surface"
             if not overwrite_run_dir.exists():
                 available = [x for x in premade_rundir_path.iterdir() if x.is_dir()]
@@ -1478,6 +1478,8 @@ class experiment:
                     f"Surface forcing {surface_forcing} not available. Please choose from {str(available)}"  ##Here print all available run directories
                 )
         else:
+            ## In this case there's no separate forcing so don't need to alter the run dir.
+            #! When implementing tides we need an extra elif statement, as we'll need to overwrite files in rundir to include tides!
             overwrite_run_dir = False
 
         # 3 different cases to handle:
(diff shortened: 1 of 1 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -1470,14 +1470,16 @@ def setup_run_directory(
...
+        if type(surface_forcing) == str:
             overwrite_run_dir = premade_rundir_path / f"{surface_forcing}_surface"
             if not overwrite_run_dir.exists():
                 available = [x for x in premade_rundir_path.iterdir() if x.is_dir()]
                 raise ValueError(
                     f"Surface forcing {surface_forcing} not available. Please choose from {str(available)}"  ##Here print all available run directories
                 )
         else:
+            ## In this case there's no separate forcing so don't need to alter the run dir.
+            #! When implementing tides we need an extra elif statement, as we'll need to overwrite files in rundir to include tides!
```
navidcy, line 1482: this comment for tides is it relevant or are you just forecasting into the future?

ashjbarnes: Just forecasting for future to make sure we don't forgoet

navidcy: OK, I see. But it's confusing here at this point, right? Let me suggest a rephrase

navidcy, line 1482:
```suggestion
            ## In case there is additional forcing (e.g., tides) then we need to modify the run dir to include the additional forcing.
```

navidcy (approved):
lgtm

regional_mom6/regional_mom6.py
```diff
@@ -1470,14 +1470,15 @@ def setup_run_directory(
             raise ValueError(
                 f"Cannot find the premade run directory files at \n{premade_rundir_path}\n. Something is not right about how the package has been installed as these files are missing!"
             )
-        if surface_forcing != False:
+        if type(surface_forcing) == str:
```
angus-g, line 1473: `isinstance(surface_forcing, str)`... or just `if surface_forcing` now, maybe use type hint of `Optional[str]` to suggest what is expected. Anyway, guess it's too late :)

navidcy: it's never too late, we can open another one ;)

navidcy: I don't understand your second suggestion, after "or just"... though

navidcy: oh... now I see... I didn't know that `if x:` would be triggered even if `x` is _not_ a boolean or 1.

angus-g: > In the context of Boolean operations, and also when expressions are used by control flow statements, the following values are interpreted as false: False, None, numeric zero of all types, and empty strings and containers (including strings, tuples, lists, dictionaries, sets and frozensets). All other values are interpreted as true.

From https://docs.python.org/3/reference/expressions.html#booleans

PR #137, opened 2024-03-29
Further enhance/clarify `setup_run_directory`

navidcy:
Continuing #135


regional_mom6/regional_mom6.py
```diff
@@ -1346,7 +1346,7 @@ def setup_run_directory(
             raise ValueError(
                 f"Cannot find the premade run directory files at \n{premade_rundir_path}\n. Something is not right about how the package has been installed as these files are missing!"
             )
-        if type(surface_forcing) == str:
+        if isinstance(surface_forcing, str):
```
angus-g, line 1349: This is the more Pythonic way of doing it, but I don't see why we don't just do `if surface_forcing`? I think if it is not a string and resolves to something crazy, you'd see it through the missing directory or `ValueError`.

regional_mom6/regional_mom6.py
```diff
@@ -1325,8 +1325,8 @@ def setup_run_directory(
         Args:
             regional_mom6_path (str): Path to the regional MOM6 source code that was cloned
                 from GitHub. Default is current path, ``'.'``.
-            surface_forcing (Optional[str, bool]): Specify the choice of surface forcing, one
-                of: ``'jra'`` or ``'era5'``. If left blank, constant fluxes will be used.
+            surface_forcing (Optional[str]): Specify the choice of surface forcing, one
+                of: ``'jra'`` or ``'era5'``. If not prescribed then constant fluxes are used.
```
angus-g, line 1329: Nice! By the way, for specifying that a value can have one of two types, you should use `Union[type_a, type_b]`

regional_mom6/regional_mom6.py
navidcy: ok, done!

regional_mom6/regional_mom6.py
navidcy: cc @ashjbarnes

ashjbarnes: Oh thanks yeah I'd forgotten about Union!

ashjbarnes (approved):
Sounds good!

PR #140, opened 2024-04-02
Adds interactive domain selection via a map in the ACCESS-OM2 demo

reillyja:
I've just added an interactive way of setting out the domain boundaries where you can just click and drag a box and it saves the associated boundary coordinates. It's not super clean but I can fix that up if its worth keeping.


ashjbarnes:
Thanks @reillyja ! This looks great - would we be able to incorporate your functions into the library itself so they can be imported from the package? I think it would be a matter of copy pasting and providing docstrings that are consistent with the format of the other functions in the package. That is if you're happy with this code becoming part of the repo of course!

This feature will be really helpful when Copernicus finishes replacing the old motu requests thing. Then we can build a new automatic data download function, incorporating these functions. For now we need to do the download by hand from their GUI while their API is down.

navidcy:
This sounds like a great feature. Does it only work with ACCESS-OM2?

ashjbarnes:
@reillyja Copernicus has finished updating their auto download feature again and I've got some scripts ready to that'll do the auto-downloading. Would you be able to incorporate your functions on Gadi into the repo so they can be built into the package?

navidcy:
@reillyja, #162 deleted this example, but the intention is to move it to cosima-recipes; see https://github.com/COSIMA/cosima-recipes/pull/338

reillyja:
> @reillyja Copernicus has finished updating their auto download feature again and I've got some scripts ready to that'll do the auto-downloading. Would you be able to incorporate your functions on Gadi into the repo so they can be built into the package?

Sorry - just saw this! Yes I can definitely look at adding this to the package. What's the timeline for getting this stuff done? i.e., can I look into this once we're on the boat @ashjbarnes? Just a bit snowed in at the moment

ashjbarnes:
@reillyja Hey John I've just made a new branch called `glorys-function` which has functions for automatically making the Glorys downloads. Probably a good time to add your widget now! That way when selecting the right boundaries with your widget at the top of the notebook, the `expt.longitude_extent` and `expt.latitude_extent` could be set automatically. Let us know if you have any questions about incorporating functions into the main `mom6-regional.py` file :)

navidcy:
@reillyja I believe this PR better fits in the cosima-recipes repo, e.g., here:
https://cosima-recipes.readthedocs.io/en/latest/Recipes/regional-mom6-forced-by-access-om2.html

navidcy:
I’m closing this, feel free to reopen tho.

PR #141, opened 2024-04-02
AK edits to paper

aekiss:
Hi @ashjbarnes, the paper looks great. This PR has are some edits for your consideration, and I have a few more comments below.

- FYI MOM5 can actually do regional modelling with open boundaries. I gather MOM6 is more capable though I don't know the details.

- There are some useful details in the commented-out sections which I think clarify what your package achieves - were they omitted due to space?

- The Repository link at the left on the first page of the PDF just goes to the JOSS page, not your github repo.

- Missing thumbnail in https://regional-mom6.readthedocs.io/en/latest/demos.html

Additional points/details that could be worth making in the paper:
- the package is available via conda-forge and pip for easy installation
- detailed documentation on readthedocs, including API reference
- pytests tests for both code and documentation to maintain quality

PR #142, opened 2024-04-02
Fixes for demos

navidcy:
I open this PR mainly to address a remark made by @aekiss in #141: that the docs don't show the thumbnail for the GLORYS+ERA5 example. I think this is because we need to push the notebook with cells evaluated _OR_ finish https://github.com/COSIMA/regional-mom6/issues/96 and have the docs evaluate the notebooks during the built stage. Am I right?

cc @angus-g, @ashjbarnes


ashjbarnes (approved):
looks good! Will need to update slightly according to changes in #147 . Namely bathymetry -> setup_bathymetry, and replace the dictionary with kwargs `x_coordinate_name` = "lon"` etc

navidcy:
> Will need to update slightly according to changes in #147 . Namely bathymetry -> setup_bathymetry, and replace the dictionary with kwargs `x_coordinate_name` = "lon"` etc

#147 should deal with updates due to changing API

navidcy:
Still the remark on thumbnail that was pointed out in #141 is outstanding.

PR #143, opened 2024-04-07
Paper cc

ChrisC28:
Hi @ashjbarnes @navidcy 
Following @aekiss and making my comments through a PR on a separate branch. 

I've made a few mostly minor modifications to the paper.md file. I've included a referece to Mike Herzfeld's work developing the original implementation of the radiation boundary conditions in MOM4, which includes a number of rules-of-thumb regarding regional domain parameters (grid resolution, time step, etc..). I also agree with @aekiss that there was a lot of useful information in the commented out sections of the paper, that, if space isn't an issue, I think could be included. 

A few brief comments
- Worth noting that MOM6, while improving some aspects of the regional modelling, has also lost some capabilities. Noteably, a number of radiative schemes are yet to be implemented on the C-grid. Not sure if we need this level of detail in the paper. 
- The major advantage of using MOM6 in regional modelling is to access the advantages of the new ocean model (ALE, better tracer advection schemes enabling larger time-steps, etc...);
- Do we need step-by-step guide to what inputs are required, and what the expected outputs are? Or is this best handled within the code itself?
- How removed from the NCI/Gadi environment do we need to be to make this work open for a general audience? 

Great work so far. I don't think you are far from having something ready to submit.


navidcy:
@ChrisC28 can you reopen the PR but towards `joss-paper` branch?

navidcy (changes requested):
please open the PR from `paper_CC` -> `joss-paper` branch

ashjbarnes:
Closed and migrated to a PR pointing to joss-paper PR to keep our git history tidy #144

PR #147, opened 2024-04-08
(0.5.0) Better `experiment.bathymetry` method

navidcy:
use kwargs

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -914,7 +914,7 @@ class experiment:
         Output is saved to the input directory of the experiment.
 
         Args:
-            bathymetry_path (str): Path to chosen bathymetry file netCDF file.
+            bathymetry_path (str): Path to the netCDF file with the bathymetry.
             coordinate_names (Dict[str, str]): Mapping of coordinate names between
                 the input and output. For example, ``{'xh': 'lon', 'yh': 'lat',
                 'elevation': 'z'}`` implies that the ``'xh'``, ``'yh'``, and
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -363,7 +363,7 @@ class experiment:
 
     Methods in this class generate the various input files needed for a MOM6
     experiment forced with open boundary conditions (OBCs). The code is agnostic
-    to the user's choice of boundary forcing, topography and surface forcing;
+    to the user's choice of boundary forcing, bathymetry and surface forcing;
     users need to prescribe what variables are all called via mapping dictionaries
     from MOM6 variable/coordinate name to the name in the input dataset.
 
@@ -893,14 +893,13 @@ class experiment:
         print("Done.")
         return
 
-    def bathymetry(
+    def setup_bathymetry(
         self,
         *,
         bathymetry_path,
         coordinate_names,
         fill_channels=False,
         minimum_layers=3,
-        make_topography=True,
         positive_down=False,
         chunks="auto",
     ):
@@ -927,9 +926,7 @@ class experiment:
             minimum_layers (Optional[int]): The minimum depth allowed
                 as an integer number of layers. The default value of ``3``
                 layers means that anything shallower than the 3rd
-                layer (as specified by the ``vcoord``) is deemed land.
-            make_topography (Optional[bool]): If ``True`` (default), the method
-                does something, else it does nothing.
+                layer (as specified by the vertical coordinate file ``vcoord.nc``) is deemed land.
             positive_down (Optional[bool]): If ``True``, it assumes that
                 bathymetry vertical coordinate is positive down. Default: ``False``.
             chunks (Optional Dict[str, str]): Chunking scheme for bathymetry, e.g.,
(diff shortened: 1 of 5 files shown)


navidcy:
@ashjbarnes, what is the purpose of `make_topography = True`?
all `experiment.bathymetry` seems to be indented into an `if make_topography:` which beats the purpose?

regional_mom6/regional_mom6.py
```diff
@@ -1100,7 +1121,7 @@ def tidy_bathymetry(
 
         topog.expand_dims("tiles", 0)
 
-        if not positivedown:
+        if not positive_down:
             ## Ensure that coordinate is positive down!
             topog["depth"] *= -1
```
navidcy, line 1126: why multiplying by -1 ensures that? there is an assumption that it wasn't already positive down which implies that the user was careful enough, right?

regional_mom6/regional_mom6.py
```diff
@@ -894,60 +895,75 @@ def rectangular_boundary(
...
+                ``'elevation'`` coordinates in MOM6 correspond to coordinates
+                ``'lon'``, ``'lat'``, and ``'z'`` in the bathymetry netCDF file
+                at ``bathymetry_path``.
             fill_channels (Optional[bool]): Whether or not to fill in
                 diagonal channels. This removes more narrow inlets,
                 but can also connect extra islands to land. Default: ``False``.
             minimum_layers (Optional[int]): The minimum depth allowed
                 as an integer number of layers. The default value of ``3``
                 layers means that anything shallower than the 3rd
                 layer (as specified by the ``vcoord``) is deemed land.
```
navidcy, line 930: what is `vcoord`?

navidcy:
@ashjbarnes question: Is this method here an old name? `make_bathymetry`?

https://github.com/COSIMA/regional-mom6/blob/4ceaf39809c6c955ec76b3b499276da5ac1845aa/regional_mom6/regional_mom6.py#L1287

regional_mom6/regional_mom6.py
ashjbarnes: No the user can provide their bathymetry without having to pre-multiply it. Passing the `positive_down` argument just specifies what the raw data is and the code adjusts by multiplying by -1 or not. gebco for example has positive values for land and neg for ocean whereas MOM6 wants depth to be positive

ashjbarnes:
> @ashjbarnes, what is the purpose of `make_topography = True`? all `experiment.bathymetry` seems to be indented into an `if make_topography:` which beats the purpose?

yeah that was an old thing! I'm fixing it now

regional_mom6/regional_mom6.py
navidcy: My point was that this implies that the user knows what “positive down” is. We can alternatively have the code here test to figure out whether the provided bathymetry needs to be multiplied w -1 or not and do that and print an info message to the user?

ashjbarnes:
what does the * do here? I'm not familiar with this notation

```
    def setup_bathymetry(
        self,
        *,
        bathymetry_path,
        coordinate_names,
        fill_channels=False,
        minimum_layers=3,
        positive_down=False,
        chunks="auto",
    ):
```

ashjbarnes (approved):
If @navidcy is happy with the changes I've made then this is good to go. I've switched to calling everything 'bathymetry', added more docstrings, removed redundant make_topog argument and switched the coordinate names dict to kwargs instead

navidcy:
> what does the * do here? I'm not familiar with this notation
> 
> ```
>     def setup_bathymetry(
>         self,
>         *,
>         bathymetry_path,
>         coordinate_names,
>         fill_channels=False,
>         minimum_layers=3,
>         positive_down=False,
>         chunks="auto",
>     ):
> ```

it splits the args from kwargs; `self` needs to be an argument (similarly it `__init__`)

navidcy:
@ashjbarnes can you give me an example of a vertical coordinate that it is `positive_down` and one that it is not?

ashjbarnes:
> @ashjbarnes, I'd like to automate this multiplication and remove the `positive_down` kwarg. I believe (but I'm not sure that it's true) that there should be a way so that the user can be completely naïve about this positive up/down situation and whatever forcing the user provide the package deals with it and converts the vertical coordinate to how MOM6 expects it to be. Could we do that?
> 
> Could you give me an example of a vertical coordinate that it is `positive_down` and one that it is not?

That's what I'd hoped originally, but what ended up with me accidentally inverting the bathymetry! If the user gives the bathymetry for an existing MOM6 run, then it will have depths of 0 -> 6000. If you give it GEBCO it'll have depths of -6000-> 0 and positive values correspond to land. What do you think is a robust way for the model to assume correctly the orientation of the vertical coordinate?

navidcy:
The vertical coordinate could be one of:

* [0, 10, ..., 6000] (eg MOM6)
* [0, -10, ..., -6000] (eg GEBCO)
* [-6000, ..., -10, 0] (eg Oceananigans)
* [6000, ..., 10, 0] (I have no example in mind)

Whatever the vertical coordinate convention is we need to convert it to MOM6 convention, right?

navidcy:
```python
import numpy as np

def convert_to_MOM6(vcoord):

    vcoord = np.array(vcoord)

    if len(vcoord)==1:
        return vcoord
    
    if vcoord[1] > vcoord[0]:
        ascending = True
    elif vcoord[1] < vcoord[0]:
        ascending = False
    else:
        raise ValueError("weird... vcoord seems constant?")

    if vcoord[1] > 0 & vcoord[0] >= 0:
        is_positive = True
    elif vcoord[1] < 0 & vcoord[0] <= 0:
        is_positive = False
    else:
        raise ValueError("weird... vcoord is both positive and negative?")

    if (is_positive == True & ascending == True):
        return vcoord

    if is_positive == False:
        return convert_to_MOM6(-1*vcoord)

    if ascending == False:
        return np.flip(vcoord)
    elif ascending == True:
        return vcoord


vcoords = ([0, 1, 2, 10],
           [0, -1, -2, -10],
           [10, 2, 1, 0],
           [-10, -2, -1, 0],
           np.array([0, 1, 2, 10]),
           np.array([0, -1, -2, -10]),
           np.array([10, 2, 1, 0]),
           np.array([-10, -2, -1, 0])
)

for vcoord in vcoords:
    print(convert_to_MOM6(vcoord))
```

navidcy:
In MOM6 what values does land have?

navidcy:
I decided to merge this and potentially deal with the `positive_down` kwarg and its potential automation in a different PR.

ashjbarnes:
> @ashjbarnes what do you think of the following?
> 
> ```python
> import numpy as np
> 
> def convert_to_MOM6_convention(vcoord):
>     '''
>     Converts the vertical coordinate (``vcoord``) to have the convention expected by
>     MOM6, that is positive values that depict the distance below sea level and with
>     ascending order. For example ``vcoord = [0, 10, 50, 100]``.
>     '''
> 
>     # for single vertical layer config don't do anything
>     if len(vcoord)==1:
>         return vcoord
> 
>     # ensure vcoord is numpy.array
>     vcoord = np.array(vcoord)
>     
>     # if in descending order then flip them
>     if vcoord[1] > vcoord[0]:
>         ascending = True
>     elif vcoord[1] < vcoord[0]:
>         return convert_to_MOM6_convention(np.flip(vcoord))
>     else:
>         raise ValueError("vcoord seems constant...")
> 
>     # if values are negative, convert to positive
>     if vcoord[1] > 0 and vcoord[0] >= 0:
>         is_positive = True
>     elif vcoord[1] < 0 and vcoord[0] <= 0:
>         return convert_to_MOM6_convention(-1 * vcoord)
>     else:
>         raise ValueError("vcoord has both positive and negative values...")
> 
>     # if vcoord follows MOM6 convention then all is well
>     if is_positive and ascending:
>         return vcoord
> ```
> 
> When I try it out:
> 
> ```python
> # test that convert_to_MOM6 does what is supposed to do
> vcoords = ([0, 1, 2, 10],
>            [0, -1, -2, -10],
>            [10, 2, 1, 0],
>            [-10, -2, -1, 0],
>            np.array([0, 1, 2, 10]),
>            np.array([0, -1, -2, -10]),
>            np.array([10, 2, 1, 0]),
>            np.array([-10, -2, -1, 0]),
>            [0.2, 1, 2, 10],
>            [-0.2, -1, -2, -10],
>            [10, 2, 1, 0.2],
>            [-10, -2, -1, -0.2],
> )
> 
> for vcoord in vcoords:
>     print(convert_to_MOM6_convention(vcoord))
> ```
> 
> gives
> 
> ```python
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0  1  2 10]
> [ 0.2  1.   2.  10. ]
> [ 0.2  1.   2.  10. ]
> [ 0.2  1.   2.  10. ]
> [ 0.2  1.   2.  10. ]
> ```
> 
> demonstrating that it _does_ return things how MOM6 expect them?

Gebco is more like [-5000...0...5000] How do we automatically tell which bits are land and which are ocean? That's the issue. I also don't think that whether it's 0,1,2,3 or 3,2,1,0 matters since it gets interpolated onto the base grid anyway

navidcy:
Let's continue the discussion about potential elimination of `positive_down` kwarg in #150

PR #158, opened 2024-04-22
Tweak API docstrings/comments

aekiss:


regional_mom6/regional_mom6.py
```diff
@@ -1713,9 +1717,8 @@ def __init__(
 
     def rectangular_brushcut(self):
         """
-        Cut out and interpolates tracers. This method assumes that the boundary
-        is a simple Northern, Southern, Eastern, or Western boundary. Cuts out
-        and interpolates tracers.
+        Cut out and interpolate tracers. This method assumes that the boundary
+        is a simple Northern, Southern, Eastern, or Western boundary.
         """
         if self.orientation == "north":
```
aekiss, line 1723: Should this sequence of `if` statements be `elif'`s, with a final `else` to abort on unknown orientation?

(Or have an assert that orientation is one of allowed options?

aekiss: Similarly with `if self.grid` sequence below

navidcy (approved):
great work!
lgtm!

regional_mom6/regional_mom6.py
navidcy: I like this suggestion.
@ashjbarnes, what do you think?

navidcy:
I took the liberty to push a few tweaks...

regional_mom6/regional_mom6.py
navidcy: done via [81813aa](https://github.com/COSIMA/regional-mom6/pull/158/commits/81813aa177bebe56871bafc3228c84424662b7e3) in a slightly different way than what @aekiss suggested

PR #164, opened 2024-04-23
Fix error msgs related to initial condition + add tests for ValueErrors

navidcy:


regional_mom6/regional_mom6.py
```diff
@@ -592,28 +592,29 @@ def initial_condition(
...
-        # Separate out tracers from two velocity fields of IC
+        # Separate out IC's tracers, two velocities, and free surface
         try:
             ic_raw_tracers = ic_raw[
                 [varnames["tracers"][i] for i in varnames["tracers"]]
             ]
         except:
             raise ValueError(
-                "Error in reading in initial condition tracers. Terminating!"
+                "Error while reading initial conditions tracers. Terminating!"
```
angus-g, line 602:
```suggestion
                "Error while reading initial condition tracers. Terminating!"
```

codecov.yml
```diff
@@ -6,6 +6,7 @@ codecov:
 comment: false
 
 ignore:
+  - "3/*"
```
angus-g, line 9: What is this?

tests/test_expt_class.py
```diff
@@ -160,80 +160,132 @@ def test_ocean_forcing(
...
+        np.random.random((100, 100, 10)),
+        dims=dims,
+        coords=coords,
+    )
+
+    v = xr.DataArray(
+        np.random.random((100, 100, 10)),
+        dims=dims,
+        coords=coords,
+    )
```
angus-g, line 201: Probably clearer with something like:

```python
def rand_ds():
   return xr.DataArray(
        np.random.random((100, 100, 10)),
        dims=dims,
        coords=coords,
    )

ic_datasets = {var: rand_ds() for var in ["temp", "salt", "u", "v"]}
```

tests/test_expt_class.py
```diff
@@ -160,80 +160,132 @@ def test_ocean_forcing(
...
+            "v": v,
+        }
+    )
+
+    initial_cond_without_v = xr.Dataset(
+        {
+            "temp": temp,
+            "salt": salt,
+            "eta": eta,
+            "u": u,
```
angus-g, line 265: Maybe like...
```python
def ic_dataset_without(ics, keys):
    return xr.Dataset(
        {k: v for k, v in ics.items() if k not in keys}
    )
```

angus-g (commented):
I think there are a few tweaks to the test to reduce duplication/increase clarity, but I do like the extended coverage by adding all the new conditions!

codecov.yml
navidcy: Sorry, I should have explained.
That's my attempt to remove this `3` directory from the codecov report, e.g., see 

Screenshot 2024-04-26 at 07 16 12

I have no idea why is there! Does it come from a docker/CI on the notebook?

tests/test_expt_class.py
navidcy: Thanks for the suggestions! I would really like to reduce the boilerplate code but I didn't have the python-expertise to do it from the top of my head!

navidcy:
this is WIP -- I think I know how to proceed just didn't find the time!

aekiss:
maybe convert to a draft PR to avoid confusion?

navidcy:
great idea!

navidcy:
I'm closing this as stale :(

PR #167, opened 2024-04-24
fix regional-mom6 link in docs/index.rst

aekiss:


aekiss:
fix md-style link in rst file.

(why are the doc files a mix of md and rst?)

navidcy:
for historical reasons: @angus-g started it with rat but I'm more comfortable with Markdown, so that's why...
we should homogenize it at some point, but it's low priority...

navidcy (approved):

PR #170, opened 2024-04-30
Era5 fix. Typo, and need to combine years into single file again

ashjbarnes:
I'd separated ERA5 forcing files into separate years automatically, but this probably isn't the best workflow. If someone does do an IAF run with HUGE forcing files, they should be able to split the forcing files up yearly as needed themselves. As it stands the package tried to anticipate this and split each ERA5 field into each forcing year, but this requires extra overhead and explaining unnecessary for simple runs.

1. Fix typo
2. Simplify ERA5 forcing files so that there's one file per field rather than per field and year.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1572,85 +1572,94 @@ class experiment:
             years = [
                 i for i in range(self.date_range[0].year, self.date_range[1].year + 1)
             ]
-            # Loop through each year and read the corresponding files
-            for year in years:
-                ds = xr.open_mfdataset(
-                    f"{era5_path}/{fname}/{year}/{fname}*",
-                    decode_times=False,
-                    chunks={"longitude": 100, "latitude": 100},
+            # construct a list of all paths for all years to use for open_mfdataset
+            paths_per_year = [Path(f"{era5_path}/{fname}/{year}/") for i in years]
+            all_files = []
+            for path in paths_per_year:
+                # Use glob to find all files that match the pattern
+                files = list(path.glob(f'{fname}*.nc'))
+                # Add the files to the all_files list
+                all_files.extend(files)
+
+            ds = xr.open_mfdataset(
+                all_files,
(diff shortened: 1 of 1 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -1572,85 +1572,86 @@ def setup_era5(self, era5_path):
...
-                )
+            ## Cut out this variable to our domain size
+            rawdata[fname] = longitude_slicer(
+                ds,
+                self.longitude_extent,
+                "longitude",
+            ).sel(
+                latitude=slice(
+                    self.latitude_extent[1], self.latitude_extent[0]
+                )  ## This is because ERA5 has latitude in decreasing order (??)
```
navidcy, line 1598: What is this question here? Can’t we find out if that’s true or not and remove the “(??)”?

navidcy (commented):
I’ll have a closer look later. 
could we discuss about how to have a test or how to make the downloading more robust?

regional_mom6/regional_mom6.py
ashjbarnes: I'm not really sure tbh - might have been a reason for this but I never wrote it down. Don't have too much time today to troubleshoot but was just trying to get this patched for Anupiya (new student of Callum) who couldn't get the example working

ashjbarnes: Need to make the most of Gadi being up before it takes a 2 day holiday!!

ashjbarnes:
The downloading? We currently don't do any automatic downloading because the Copernicus Marine client is still very unstable and undergoing work. this package currently assumes you've already done all the downloading

ashjbarnes:
This patch doesn't need to be perfect for now, but until it goes through the example won't work when running MOM6

navidcy:
Anupiya can install from this branch. We should merge this when properly reviewed

navidcy (approved):

Issue #174, opened 2024-05-08
Iterate through boundaries as part of `expt.initial_condition` method


navidcy:
Currently the [demo](https://regional-mom6.readthedocs.io/en/latest/demo_notebooks/reanalysis-forced.html#Step-5:-Handle-the-ocean-forcing---where-the-magic-happens) has

```python
# Set up the initial condition
expt.initial_condition(
    glorys_path / "ic_unprocessed.nc", # directory where the unprocessed initial condition is stored, as defined earlier
    ocean_varnames,
    arakawa_grid="A"
    )

# Now iterate through our four boundaries
for i, orientation in enumerate(["south", "north", "west", "east"]):
    expt.rectangular_boundary(
        glorys_path / (orientation + "_unprocessed.nc"),
        ocean_varnames,
        orientation,    # Needs to know the cardinal direction of the boundary
        i + 1,          # Just a number to identify the boundary. Indexes from 1
        arakawa_grid="A"
        )
```

Given that most times most users will want to create all 4 boundaries, I am wondering if we can make the `for i, orientation in enumerate(["south", "north", "west", "east"])` loop part of the `expt.initial_condition` method. Perhaps we can have a default kwarg which will be `["south", "north", "west", "east"]` and if the user wants to only do one of the boundaries they can provide a different value for the kwarg?

Something like that will simplify the demo.

ashjbarnes:
This is how things used to be actually but I changed it to make things more flexible. See #94

ashjbarnes:
I agree that it would make the IC easier to parse though! Maybe there's a middle ground. I decided to split them because I also thought that adding too many kwargs and options enabling "IC with boundary" or "boundary without IC" would be confusing too

navidcy:
> I agree that it would make the IC easier to parse though! Maybe there's a middle ground. I decided to split them because I also thought that adding too many kwargs and options enabling "IC with boundary" or "boundary without IC" would be confusing too

A kwarg with default would not required to be provided. Users would only provide it if they need something else other than the default.

navidcy:
As is now it’s a bit cumbersome.

ashjbarnes:
@angus-g what do you think? I think this PR #94 we'd chatted about over coffee but discussion didn't make it to github

angus-g:
> Perhaps we can have a default kwarg which will be ["south", "north", "west", "east"] and if the user wants to only do one of the boundaries they can provide a different value for the kwarg?

[...]

PR #175, opened 2024-05-09
(v0.6.0) Simplify boundary condition construction

navidcy:
*Description edited by Ashley to reflect discussion below*
This PR aims to simplify the boundary condition construction in the simple case of four boundaries that are parallel to lines of constant longitude and latitude, as per the discussion in https://github.com/COSIMA/regional-mom6/issues/174.

To do this, we add a wrapper function that just calls the `simple_boundary` method for each cardinal direction requested. 

This PR includes a breaking API change since `rectangular_boundary` is renamed to `rectangular_boundaries` and also some of its args changed.

Closes https://github.com/COSIMA/regional-mom6/issues/174

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -890,7 +890,6 @@ class experiment:
             },
         )
 
-
         self.ic_eta = eta_out
         self.ic_tracers = tracers_out
         self.ic_vels = vel_out
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -568,18 +568,17 @@ class experiment:
 
     def initial_condition(
         self,
-        ic_path,
+        raw_ic_path,
         varnames,
         arakawa_grid="A",
         vcoord_type="height",
-        boundaries=["south", "north", "west", "east"],
     ):
         """
         Reads the initial condition from files in ``ic_path``, interpolates to the
         model grid, fixes up metadata, and saves back to the input directory.
 
         Args:
-            ic_path (Union[str, Path]): Path to initial condition file.
+            raw_ic_path (Union[str, Path]): Path to raw initial condition file to read in.
             varnames (Dict[str, str]): Mapping from MOM6 variable/coordinate names to the names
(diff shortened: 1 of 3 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -874,11 +879,23 @@ def initial_condition(
...
-        print("done setting up initial condition.")
 
         self.ic_eta = eta_out
         self.ic_tracers = tracers_out
         self.ic_vels = vel_out
+
+        # Now iterate through our four boundaries
+        for i, orientation in enumerate(boundaries, start=1):
+            self.rectangular_boundary(
+                glorys_path / (orientation + "_unprocessed.nc"),
```
navidcy, line 890: what this should be? definitely not `glorys_path` right?

ashjbarnes:
In this PR you're changing the boundary condition function not the initial condition right? I still think that the boundary and initial conditions should be kept as separate functions. If not, then we'll should rename it so that the "initial condition" function doesn't do the boundary right?

regional_mom6/regional_mom6.py
ashjbarnes: You could assume that the user has put the boundary forcing files in the same folder as the initial condition I guess? This is the path that's passed to the 'initial condition' function

navidcy:
Oh my bad

ashjbarnes:
This PR will require a new tag release as it's not backwards compatible. Also need to update [the access om2 example](https://github.com/COSIMA/cosima-recipes/pull/338) with the changes.

ashjbarnes:
I'm going to approve the PR because it was Navid's originally. Wait on merging until @navidcy is happy and we also fix up the access om2 example to match simultaneously

ashjbarnes (approved):

navidcy:
I'd like to merge this and then we tag a new release -- how does this sound @ashjbarnes?

ashjbarnes:
Great! Need to reflect changes in [access om2 example](https://github.com/COSIMA/cosima-recipes/pull/338)

PR #177, opened 2024-06-11
Make setup rundir more robust

ashjbarnes:
The `setup_run_directory` method has two problems:

1. It assumes that the package has been installed via conda when it looks for the `premade_run_directories` path.
2. If reading in an existing mask table and the user hasn't run the FRE tools method, it crashes because the cpu layout hasn't been defined anywhere in this case.

These issues are addressed via some simple if-statement error handing.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1425,6 +1425,18 @@ class experiment:
         premade_rundir_path = Path(
             importlib.resources.files("regional_mom6") / "demos/premade_run_directories"
         )
+        if not premade_rundir_path.exists():
+            print("Could not find premade run directories at ", premade_rundir_path)
+            print("Perhaps the package was imported directly rather than installed with conda. Checking if this is the case... ")
+        
+            premade_rundir_path = Path(
+                importlib.resources.files("regional_mom6").parent() / "demos/premade_run_directories"
+            )
+            if not premade_rundir_path.exists():
+                raise ValueError(
+                    f"Cannot find the premade run directory files at {premade_rundir_path} either.\n\n"
+                    + "There may be an issue with package installation. Check that the `premade_run_directory` folder is present in one of these two locations"
+                )
 
         # Define the locations of the directories we'll copy files across from. Base contains most of the files, and overwrite replaces files in the base directory.
         base_run_dir = premade_rundir_path / "common_files"
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -437,7 +437,7 @@ class experiment:
         self.grid_type = grid_type
         self.repeat_year_forcing = repeat_year_forcing
         self.ocean_mask = None
-        self.layout = None # This should be a tuple. Leaving in a dummy 'None' makes it easy to remind the user to provide a value later on.
+        self.layout = None  # This should be a tuple. Leaving in a dummy 'None' makes it easy to remind the user to provide a value later on.
         if read_existing_grids:
             try:
                 self.hgrid = xr.open_dataset(self.mom_input_dir / "hgrid.nc")
@@ -1428,10 +1428,13 @@ class experiment:
         )
         if not premade_rundir_path.exists():
             print("Could not find premade run directories at ", premade_rundir_path)
-            print("Perhaps the package was imported directly rather than installed with conda. Checking if this is the case... ")
-        
+            print(
+                "Perhaps the package was imported directly rather than installed with conda. Checking if this is the case... "
+            )
+
             premade_rundir_path = Path(
-                importlib.resources.files("regional_mom6").parent / "demos/premade_run_directories"
+                importlib.resources.files("regional_mom6").parent
+                / "demos/premade_run_directories"
             )
             if not premade_rundir_path.exists():
                 raise ValueError(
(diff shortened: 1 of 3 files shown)


navidcy:
I'd feel more comfortable if @angus-g had a look here.

ashjbarnes:
To be more verbose:

If you install from conda, then `importlib.resources.files("regional_mom6") / "demos/premade_run_directories"` lands you in the right place, namely:
`pathtocode/regional_mom6/demos/premade_run_directories`
However, if you import it directly then you end up one folder too deep, I.E 
`pathtocode/regional_mom6/regional_mom6/demos/premade_run_directories`

Hence with my fix I just go back one directory if we detect that the path doesn't exist. 

For the second point, I've made a default placeholder for the `layout` attribute so that we don't end up calling `self.attribute` when it doesn't exist. There are probably more elegant ways of doing this but this works.

regional_mom6/regional_mom6.py
```diff
@@ -437,6 +437,7 @@ def __init__(
         self.grid_type = grid_type
         self.repeat_year_forcing = repeat_year_forcing
         self.ocean_mask = None
+        self.layout = None  # This should be a tuple. Leaving in a dummy 'None' makes it easy to remind the user to provide a value later on.
```
navidcy, line 440: how does making a default `None` reminds the user? the user doesn't see the source code -- right?

ashjbarnes: Look at the changes in the setup rundir function. There it checks whether layout is none

navidcy: great!

angus-g:
Honestly I think the better approach here would be to move `demos` under `regional_mom6`, which would also align with the PyPA recommendation (https://setuptools.pypa.io/en/latest/userguide/datafiles.html#non-package-data-files):

> Instead, the PyPA recommends that any data files you wish to be accessible at run time be included inside the package.

That just requires an additional modification to the `pyproject.toml`:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index e427db4..beee91a 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,9 +26,6 @@ build-backend = "setuptools.build_meta"
 [tool.setuptools]
 packages = ["regional_mom6", "regional_mom6.demos.premade_run_directories"]
 
-[tool.setuptools.package-dir]
-"regional_mom6.demos" = "demos"
-
 [tool.setuptools.package-data]
 "regional_mom6.demos.premade_run_directories" = ["**/*"]
 
```

navidcy:
Let's do that then? New PR -- this one?

ashjbarnes:
Ok cool good to know! I need to learn more about these conventions rather than making it up all the time.

so this way it will import correctly whether installed via conda or imported directly from a folder (as might be the case for testing)

angus-g:
> so this way it will import correctly whether installed via conda or imported directly from a folder (as might be the case for testing)

That's right, although it's also better for testing to do an editable install with `pip install -e` rather than importing the directory itself.

ashjbarnes:
> > so this way it will import correctly whether installed via conda or imported directly from a folder (as might be the case for testing)
> 
> That's right, although it's also better for testing to do an editable install with `pip install -e` rather than importing the directory itself.

even with Conda? That's the preferred install now right?

angus-g:
> even with Conda? That's the preferred install now right?

For actually using the package, yes, but we're talking about the development case. I would suggest the best way for development is to create a conda environment for regional MOM6 (or use an existing one with all the right dependencies in place), and then use an editable install of your working copy of the repository into that environment. Then any changes within the repository are immediately available without needing any additional install, or modifications to path.

navidcy:
How should we move forward for this?

ashjbarnes:
For now I'd be happy with merging this PR as is (since it does address the two issues I outline in the description, however clunkily), and making an issue pointing out that a cleaner implementation would be changing the folder location and doing whatever needs to be done on the Conda side. It's a bit of an edge case issue so I'm not going to prioritise implementing a 'clean' solution right now

navidcy:
@angus-g?

angus-g:
I would remove the special-casing code (i.e. the added lines 1429-1447) and perhaps make a documentation suggestion for development to use editable installs.

ashjbarnes:
I've got a headache because I tried to change the directory structure to match @angus-g 's suggestion. However, testing failed before it even needed the premade run directories! I force reset to two different commits that _previously_ passed the tests, and they now throw this new error. Is it possible that our tests / code aren't frozen to specific versions, and so will change over time? 

The current commit passed pytest 3 weeks ago. I guess the evidence of that is now hidden by my overwriting of the git history by forcing the head back in time...

angus-g:
The *answers* are the same, but retriggering the test has pulled in numpy 2.0.0, which has updated in the meantime. The only difference is that it's being more strict about the expected types in the test. I can quickly fix that and you can rebase on there.

ashjbarnes:
I also wasn't sure how to update the `.toml` file when moving `premade_run_dirs` from `~/demos_/` to `~/regional_mom6/`. For instance, do we still need to explicitly mention the path to these files as `package-data` in 

```[tool.setuptools]
packages = ["regional_mom6", "regional_mom6.demos.premade_run_directories"]

[tool.setuptools.package-dir]
"regional_mom6.demos" = "demos"

[tool.setuptools.package-data]
"regional_mom6.demos.premade_run_directories" = ["**/*"]
```

navidcy:
a test broke?

angus-g:
Yes, we still need it as package data. We essentially just stop "remapping" where the demos directory is. The only modification you need is here: https://github.com/COSIMA/regional-mom6/pull/177#issuecomment-2164129985

regional_mom6/regional_mom6.py
```diff
@@ -198,9 +198,9 @@ def hyperbolictan_thickness_profile(nlayers, ratio, total_depth):
                54.91569316, 63.76993631, 70.3615673 , 74.6060059 , 77.08936249,
                78.46160957, 79.19600916, 79.58232451, 79.7836947 , 79.88816229])
         >>> dz.sum()
-        1000.0
+        np.float64(1000.0)
         >>> dz[-1] / dz[0]
-        3.9721960481753706
+        np.float64(3.9721960481753706)
```
navidcy, line 203: what is this `np.float64` thing?

navidcy: it's elsewhere also....

angus-g: That was required as part of the numpy 2.0.0 compatibility to make the doctest pass: it prints out the types of return values. I think that is useful, but given we're not going to use numpy 2.0.0 for a little bit, it'll disappear for now anyway!

angus-g:
It's because the docstring changes that I was experimenting got merged in here. In the end, we needed to pin numpy < 2.0.0, so the changes to docstrings can be reverted.

navidcy:
@ashjbarnes can you open an issue linking to the discussion in this PR regarding the improved method that @angus-g suggested so we don't forget about it?

PR #178, opened 2024-07-01
Pin numpy<2.0.0

angus-g:
~~I guess it's nice that we have type information encoded here now too. This does introduce a bit of an incompatibility *in the testing environment*, where numpy 2.0.0 is required, but it's not an issue for actual running.~~

Famous last words: it *is* an issue for running, at the moment. We will probably just have to restrict ourselves to numpy < 2.0.0 for the moment.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -23,7 +23,7 @@ jobs:
         shell: bash -el {0}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -23,7 +23,7 @@ jobs:
         shell: bash -el {0}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


angus-g:
I think the new error we're seeing is an incompatibility between xESMF and numpy 2.0.0 (maybe):

```
ValueError: cannot reshape array of size 161438054400 into shape (np.int32(249), np.int32(140), np.int32(2412), np.int32(1920))
```

32-bit ints are not big enough to represent the products of the dimensions, and I wonder if the overflow there is causing it to fail a check somewhere:

```
>>> import numpy as np
>>> np.int32(249)*np.int32(140)*np.int32(2412)*np.int32(1920)
<stdin>:1: RuntimeWarning: overflow encountered in scalar multiply
-1770702848
```

The fix for the moment might be to just pin numpy to <2.0.0.

ashjbarnes:
Nice! Ok should we merge this right away and rebase all active PRs back to `main`?

ashjbarnes (approved):

angus-g:
How odd that it's the only one to hang!

navidcy:
I'm happy to just remove the 3.11 tbh...

PR #179, opened 2024-07-01
Make setup rundir more robust take #2

ashjbarnes:
This is a clone of #177 with three cherry picked commits. The other PR got really messy as @angus-g needed to handle the update to numpy 2 in between

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1425,6 +1425,18 @@ class experiment:
         premade_rundir_path = Path(
             importlib.resources.files("regional_mom6") / "demos/premade_run_directories"
         )
+        if not premade_rundir_path.exists():
+            print("Could not find premade run directories at ", premade_rundir_path)
+            print("Perhaps the package was imported directly rather than installed with conda. Checking if this is the case... ")
+        
+            premade_rundir_path = Path(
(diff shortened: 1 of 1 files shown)


angus-g:
There's nothing wrong with force pushing to your PR branches. You could've essentially just pushed this branch to the one for #177. :)

ashjbarnes:
In this PR I gave up on changing structure to move the `demos` folder to `regional-mom6`, because it required altering the `pyproject.toml` and then troubleshooting the github workflow file. To implement this in line wiht conda guidelines (see  [Angus' comment](https://github.com/COSIMA/regional-mom6/pull/177#issuecomment-2164129985) ) this will take a bit more fiddling around, and another PR. 

What we have in this PR means that there's extra error handling when the CPU layout isn't provided, and the `setup_rundir` function looks at the package root folder for the `premade_run_directories` if it can't find the folder where it expects it, in the case that the user is being lazy and importing it directly for development purposes.

navidcy (approved):

ashjbarnes:
> There's nothing wrong with force pushing to your PR branches. You could've essentially just pushed this branch to the one for #177. :)

Oh thanks good to know! I thought I'd made a huge mess of the other one by force pushing like 4 times either side of your numpy PR

PR #180, opened 2024-08-21
added glorys download functions with new API

ashjbarnes:
It seems like the new Copernicus API is working. I've re-written the functions that automatically generate a bash script for downloading the right files, incorporated into the experiment class and the example notebook

### DIFF
--- a/demos/reanalysis-forced.ipynb
+++ b/demos/reanalysis-forced.ipynb
@@ -117,22 +117,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Step 2: Prepare ocean forcing data\n",
-    "\n",
-    "We need to cut out our ocean forcing. The package expects an initial condition and one time-dependent segment per non-land boundary. Naming convention is `\"east_unprocessed\"` for segments and `\"ic_unprocessed\"` for the initial condition.\n",
-    "\n",
-    "Data can be downloaded directly from the [Copernicus Marine data store](https://data.marine.copernicus.eu/product/GLOBAL_MULTIYEAR_PHY_001_030/download) via their GUI (once logged in).\n",
-    "\n",
-    "1. Initial condition: Using the GUI, select an area matching your `longitude_extent` and `latitude_extent` that corresponds to the first day in your date range. Download the initial condition and save it with filename `ic_unprocessed.nc` inside the `glorys_path` directory.\n",
-    "2. Boundary forcing: Using the GUI, select the Eastern boundary of your domain (if you have one that contains ocean). Allow for a buffer of ~0.5 degrees in all directions, and download for the prescribed `date_range`. Download and name `east_unprocessed.nc`.\n",
-    "3. Repeat step 2 for the remaining sections of the domain."
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "## Step 3: Make experiment object\n",
+    "## Step 2: Make experiment object\n",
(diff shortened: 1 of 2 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -899,6 +945,68 @@ def initial_condition(
...
+                self.date_range[0] + datetime.timedelta(days=1),
+            ],
+            "ic_unprocessed",
+            raw_boundaries_path,
+            modify_existing=False,
+        )
+        if "east" in boundaries:
+            get_glorys_data(
+                [self.longitude_extent[1] - 1, self.longitude_extent[1] + 1],
+                [self.latitude_extent[0] - 1, self.latitude_extent[1] + 1],
```
navidcy, line 975: what is `+1` or `-1`? buffer? hardcoded?
perhaps add a comment explaining why we need that and why 1 is enough?

regional_mom6/regional_mom6.py
navidcy, line 1: anyway we can have a test for these two new methods?

demos/reanalysis-forced.ipynb
```diff
@@ -117,22 +117,7 @@
    "cell_type": "markdown",
```
navidcy, line 117: <p>is <code>get_glorys_data</code> a method of the <code>experiment</code> object of a method of <code>rmom6</code>? the text suggest the latter and the code below suggest the former.</p><p> </p><p>we can even drop the <code>boundaries</code> kwarg below since we are using the default? would that be simpler? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/COSIMA/regional-mom6/pull/180/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

navidcy:
I added few comments but not really a review.

demos/reanalysis-forced.ipynb
ashjbarnes: <p>there are two functions. One is a method, one is a standalone function that does one file at a time</p><p> </p><p>No, dropping the boundaries kwarg would then mean that the simple use case of having all land on one or more side would no longer be supported by this method</p>

regional_mom6/regional_mom6.py
ashjbarnes: Glorys is 12th degree, so 1 degree buffer is more than enough space. This method is only for glorys so it's fine for it to be hardcoded

demos/reanalysis-forced.ipynb
ashjbarnes: In fact I think 1 degree is overkill - I'll reduce it to save download space and time

navidcy: Great!

regional_mom6/regional_mom6.py
navidcy: OK great. But just for the records I wasn't commenting whether it's an overkill or not. I was just thinking that a comment there might help future developers who eg are trying to modify this methods or apply it to a different dataset. All good.

regional_mom6/regional_mom6.py
navidcy: In fact on second thought the `get_glorys_rectangular` method is called in the notebook which is run as part of the CI. And `get_glorys_rectangular` calls upon `get_glorys_data` so actually both are tested.

navidcy (approved):
lgtm!

navidcy:
After this is merged we might wanna make a patch release to 0.6.1.

regional_mom6/regional_mom6.py
ashjbarnes: Yeah makes sense! I added a comment too :)

navidcy:
I pushed a new patch release... It should be 1-2 days before it's available in conda

Issue #119, opened 2024-02-29
Testing the demo notebooks using the docker image locally


navidcy:
I don't know how can I test the demo notebooks using the docker image locally. That is, I'd like to be able to reproduce the tests that the GitHub Actions do locally before I push.

Ideally we should update the [Contributors section](https://regional-mom6.readthedocs.io/en/latest/contributing.html#testing) in the documentation to include these instructions.

ashjbarnes:
Yeah this would be good! I've learned recently how to set up local docker envs so can have a go at this.

navidcy:
cc @angus-g

ashjbarnes:
I had a go using [act](https://github.com/nektos/act) which is meant to emulate github actions but unfortunately it doesn't work that well out of the box. Formatting was failing because it couldn't install black. I think it's not worth spending the time to troubleshoot!

I can have a go directly with Docker, but in my experience this then requires some fiddling around to get it to work on different machines so could add a bit of a burden of documentation and maintenance

navidcy:
I think I'd like to resolve this before we submit to JOSS (#100)

angus-g:
Yep, I'll take a look at this. I have a few ideas, I'll figure out how feasible and easy to implement they are.

angus-g:
Okay, I've managed to get act working, which I think is probably the most consistent way of doing things. Here's a summary of what I did, maybe you can see if you can replicate, and then we can write that up.

1. Install Docker, then install [act](https://github.com/nektos/act). For my setup I did `gh extension install https://github.com/nektos/gh-act`
2. Create `event.json` containing the following:
```json
{
  "act": true
}
```
3. Use the following modification to `testing.yml` (which we can add to the repository version if this seems like the way to go). These changes only alter the workflow when we're running within act. Essentially it skips the formatting step (just run black locally...), and because our testing image doesn't include Node, it needs to install that so that the step for installing Conda works. Again, if this is the way we end up going, we should just include Node in the image itself.
```diff
diff --git a/.github/workflows/testing.yml b/.github/workflows/testing.yml
index f406515..d1da3c1 100644
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -11,6 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: psf/black@stable
+        if: ${{ !github.event.act }}
         with:
           src: "./regional_mom6 ./tests"
 
[...]

PR #183, opened 2024-08-23
Add some diagnostic plots + enhance phrasing in Demo notebook

navidcy:
I added some plots in the notebook. But I have forgotten how to actually run this notebook so that the plots appear and save them...

@angus-g or @ashjbarnes could you remind me how to do that?

### DIFF
--- a/demos/reanalysis-forced.ipynb
+++ b/demos/reanalysis-forced.ipynb
@@ -51,6 +51,7 @@
     "import regional_mom6 as rmom6\n",
     "\n",
     "import os\n",
+    "import matplotlib.pyplot as plt\n",
     "from pathlib import Path\n",
     "from dask.distributed import Client"
    ]
--- a/demos/reanalysis-forced.ipynb
+++ b/demos/reanalysis-forced.ipynb
@@ -25,7 +25,7 @@
     "\n",
     "Setting up a regional model in MOM6 can be a pain. The goal of this package is that users should spend their debugging time fixing a model that's running and doing weird things, rather than puzzling over a model that won't even start.\n",
     "\n",
-    "In running this notebook, you'll hopefully have a running MOM6 regional model. There will still be a lot of fiddling to do with the `MOM_input` file to make sure that the parameters are set up right for your domain, and you might want to manually edit some of the input files. *But*, this package should help you bypass most of the woes of regridding, encoding and understanding the arcane arts of the MOM6 boundary segment files. "
+    "In running this notebook, you'll hopefully have a running MOM6 regional model. There will still be a lot of fiddling to do with the `MOM_input` file to make sure that the parameters are set up right for your domain, and you might want to manually edit some of the input files. *But*, this package should help you bypass most of the woes of regridding, encoding, and understanding the arcane arts of the MOM6 boundary segment files. "
    ]
   },
   {
@@ -147,13 +147,22 @@
    "source": [
     "We can now access the horizontal and vertical grid of the regional configuration via `expt.hgrid` and `expt.vgrid` respectively.\n",
     "\n",
-    "Plotting the vertical grid with `marker = '.'` lets you see the spacing. You can use `numpy.diff` to compute the vertical spacings, e.g.,\n",
-    "```python\n",
-    "import numpy as np\n",
-    "np.diff(expt.vgrid.zl).plot(marker = '.')\n",
-    "```\n",
-    "shows you the vertical spacing profile.\n",
-    "\n",
+    "Plotting the vertical grid with `marker = '.'` lets us see the spacing."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "expt.vgrid.zl.plot(marker = '.', y='zl', yincrease=False, figsize=(4, 8))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "### Modular workflow!\n",
     "\n",
     "After constructing your `expt` object, if you don't like the default `hgrid` and `vgrid` you can simply modify and then save them back into the `expt` object. However, you'll then also need to save them to disk again. For example:\n",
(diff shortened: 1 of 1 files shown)


navidcy:
I actually remembered now a whole discussion about [how to do it with the docker](https://github.com/COSIMA/regional-mom6/issues/119) which I never managed to do...

navidcy:
I see there is an issue with matplotlib. I can drop that bit and just use xarray's plot functionality. We shouldn't include matplotlib as a dependency of the regional-mom6 package but I'm not sure if we can include it in the docker env. But if that's cumbersome let's just not do that -- just let me know and I'll edit accordingly.

demos/reanalysis-forced.ipynb
```diff
@@ -147,21 +150,32 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "### Modular workflow!\n",
     "\n",
-    "After constructing your `expt` object, if you don't like the default `hgrid` and `vgrid` you can simply modify and then save them back into the `expt` object. However, you'll then also need to save them to disk again. For example:\n",
+    "After constructing our `expt` object, if we don't like default horizontal and vertical grids (`hgrid` and `vgrid`) we can modify and then save them back into the `expt` object. However, we will also need to save them to disk again. For example:\n",
```
ashjbarnes, line 171:
```suggestion
    "After constructing our `expt` object, if we don't like the default horizontal and vertical grids (`hgrid` and `vgrid`) we can modify and then save them back into the `expt` object. However, we will also need to save them to disk again. For example:\n",
```

ashjbarnes: I think it's easiest to just include matplotlib in the package. No, it's not needed by the package, but it's the first thing people will install when they make a new env and realise it's missing

navidcy: I don't think we should have matplotlib as a dependency.

But I was googling this a bit and saw that there is a way to have "optional dependencies"?

https://stackoverflow.com/questions/41268863/what-is-the-difference-between-extras-require-and-install-requires-in-se/45043494#45043494

That might be a nice way? @angus-g have you seen this before or know how it works?

navidcy: from chatGPT:

If you want to use **optional dependencies** with `setuptools` but **without** a `setup.py` file, you can still achieve this by using the **`pyproject.toml`** configuration. This is in line with the modern PEP 517 and PEP 518 standards, which define the use of `pyproject.toml` to manage build tools like `setuptools`.

Here’s how you can set it up using **`setuptools`** in `pyproject.toml`:

### **Step-by-Step Setup Using `pyproject.toml` with `setuptools`**

1. **Create or Update `pyproject.toml`**:
   
   You need to configure `setuptools` as your build backend in the `pyproject.toml` file.

   Here's an example:

   ```toml
   [build-system]
   requires = ["setuptools", "wheel"]
   build-backend = "setuptools.build_meta"

   [project]
   name = "yourpackage"
   version = "0.1.0"
   description = "A package with optional dependencies"
   authors = [
       { name = "Your Name", email = "you@example.com" }
   ]
   dependencies = [
       # Add essential dependencies here, e.g.
       "requests",
   ]

   [project.optional-dependencies]
   visualization = ["matplotlib"]
   ```

   - **`[build-system]`** specifies that you're using `setuptools` and `wheel` for your package's build system.
   - **`[project]`** defines the core metadata of your project (name, version, etc.).
   - **`dependencies`** lists the required dependencies for your package.
   - **`[project.optional-dependencies]`** is where you define optional features like `visualization` with its corresponding dependencies.

2. **Instruct Users**:
   
   After setting up `pyproject.toml`, users can install the package with optional dependencies as follows:

   ```bash
   pip install yourpackage[visualization]
   ```

   This command will install the core package plus the optional `matplotlib` dependency under the `visualization` group.

3. **Handle Conditional Imports**:

   In your code, you can use `try-except` to conditionally import the optional dependency:

   ```python
   try:
       import matplotlib.pyplot as plt
   except ImportError:
       raise ImportError("Matplotlib is required for visualization. Please install it with `pip install yourpackage[visualization]`.")
   ```

---

### **Benefits of Using `pyproject.toml` with `setuptools`**:
- **No `setup.py` needed**: `pyproject.toml` serves as the modern alternative to `setup.py` and provides a clean way to configure optional dependencies.
- **Standardized build configuration**: This method complies with PEP 517 and PEP 518, ensuring compatibility with modern Python tooling.
- **Optional dependencies in a clean format**: You can define multiple optional groups (e.g., `visualization`, `testing`, `docs`), and users can choose which extras to install.

---

### **Example for Multiple Optional Dependencies**

If you want to define multiple optional features (e.g., `visualization`, `docs`, `dev`), you can do it like this:

```toml
[project.optional-dependencies]
visualization = ["matplotlib"]
docs = ["sphinx", "myst-parser"]
dev = ["pytest", "black"]
```

Now, users can install specific extras by running:

```bash
pip install yourpackage[visualization]
# or
pip install yourpackage[dev,docs]
```

---

This setup with **`pyproject.toml`** using **`setuptools`** provides flexibility and avoids the need for `setup.py` while still giving you full control over your optional dependencies. Let me know if you need help with any specific part!

angus-g: > I don't think we should have matplotlib as a dependency.

Agreed.

> That might be a nice way? @angus-g have you seen this before? You familiar with how it works?

I have used this a bit before, but I'm not sure how it interacts with conda... I don't think it's unreasonable that somebody would already have matplotlib installed if they wanted to plot stuff anyway? Plotting isn't really the job of this package?

navidcy: hm.... probably need to have different environment yml files and mix and match... oh well..

navidcy:
We can close this PR I think

ashjbarnes:
I wouldn't want to waste the work you put into the diagnostics! Could we leave it open to remind us that it's here, and incorporate these ideas into the new example notebook that @helenmacdonald and I will work on? Then close it when we merge things together?

navidcy:
Deal

navidcy:
We can redo this later; the notebook evolved, no need to hold on the past :)

navidcy:
(The branch is always there if we need o revisit.)

PR #185, opened 2024-08-28
included a min_depth parameter and overwite this in MOM_input

### LINKED ISSUE #184: Need to override MOM6 min depth based on value set in `setup_bathymetry` function
Currently, we have a minimum number of layers set in the `setup_bathymetry` function, but this doesn't get used by MOM6. Should update the `setup_rundir` function to write a line in `MOM_override` to set `MIN_DEPTH` to the depth corresponding to the minimum number of layers chosen by the user.

ashjbarnes:
Closes #184 . See issue for details

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -485,7 +485,7 @@ class experiment:
         self.repeat_year_forcing = repeat_year_forcing
         self.ocean_mask = None
         self.layout = None  # This should be a tuple. Leaving in a dummy 'None' makes it easy to remind the user to provide a value later on.
-        self.min_depth = 0.0 # Minimum depth. Shallower water will be masked out. This value is overwritten when running "setup_bathymetry" method.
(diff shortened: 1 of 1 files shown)


regional_mom6/regional_mom6.py
```diff
@@ -1716,10 +1717,20 @@ def setup_run_directory(
 
                 if "NJGLOBAL" in lines[jj]:
                     lines[jj] = f"NJGLOBAL = {self.hgrid.ny.shape[0]//2}\n"
-
         with open(self.mom_run_dir / "MOM_layout", "w") as f:
             f.writelines(lines)
 
+        # Overwrite values pertaining to vertical structure in the MOM_input file
```
navidcy, line 1723: the new thing here is that first we `f.writelines(lines)` and then we ensure that if something was there already we overwrite it?

navidcy (approved):

PR #186, opened 2024-08-29
Tidy MOM_override file

ashjbarnes:
There's a comment at the top of the `MOM_override` file which doesn't really help anyone, and also breaks when CESM tries parsing user overrides. We're better off without it!


navidcy:
@ashjbarnes does this PR supersedes https://github.com/COSIMA/regional-mom6/pull/185? Seems that it includes all the changes proposed by https://github.com/COSIMA/regional-mom6/pull/185; should we close https://github.com/COSIMA/regional-mom6/pull/185 and focus on this?

PR #200, opened 2025-02-10
Remove boundary type argument

ashjbarnes:
In this PR I remove the `boundary_type` argument as it's deprecated and doesn't do anything anyway

I also update the example notebooks, putting the `nbval-skip` tags as tags rather than comments (prettier), and tell people to ensure they have matplotlib installed so that we don't need to comment out plotting cells


navidcy:
> I also update the example notebooks, putting the `nbval-skip` tags as tags rather than comments (prettier)

that’s nice!

PR #204, opened 2025-04-25
Rotation Test Fixes + Rotation Calculation Fixes

manishvenu:
Backstory & Details:
@navidcy created a test case to test the angle calculation in rotation.py. It's failing! Navid pointed out it's got something to do with the len_lon arg in the calculation, which goes into the modulo_around_point function we took from MOM6.

Solution/What Happened:
This is a two part error!

When we do rotation of a unit square in this test, the len_lon changes -> a unit square rotated around the origin stays constant in area but changes in the x range and y range. We can't just put 2 degrees into the function. So, I added a len_lon calculation to the test
This is an edge case, but when the value provided to `modulo_around_point` is at the very edge of the domain (as in a corner point) the modulo will jump it to the other side of the boundary! We don't want that. So, I added a (len_lon+1e-10) to make sure the edge points stay where they are.
What is modulo_around_point?
After looking at it again, modulo_around_point is supposed to return the modulo value of x in an interval [x0-(len_lon/2), x0+(len_lon/2)]. For our unit test example, where we have literally a square, each corner is literally on the edge of the range. The python implementation wasn't inclusive of the boundary, so we slightly bump len_lon to make sure that is. If this is confusing, take an example. We have x = 0.75 degrees, x0 is 0 degrees, len_lon is 1.5 degrees. The range [x0-(len_lon/2), x0+(len_lon/2)] would be [-0.75,0.75]. The modulo_around_point might make x become -0.75 because of how the previous implementation worked. The math would work out to (0.75 - (-0.75) % 1.5 + (-0.75)) = -0.75. We don't want that, points at the boundary should stay where they are. The use case for modulo_around_point is if the function operates as expected, it should map -170 degrees to the equivalent positive value in the [-180, 180] or [0, 360] range or whatever len_lon x0 selection we have.

Additional Changes:
Docs Update


manishvenu:
Hmm, I'm gonna fix Crocodile main and retry this

manishvenu:
Allright, because the diff is fine, I'm going to keep these, and we can just squash merge

manishvenu:
Cherry picked them

PR #193, opened 2024-10-11
Version 1.0: North Atlantic Update - Tides, Curved Boundaries, Config, and More!😀🤯🥳

manishvenu:
**Authors**: @ashjbarnes, @manishvenu, @helenmacdonald 

**Description**:

Ashley and I worked together to add a few features to RM6. While we get the changes ready to merge with the main COSIMA branch, please check out the list of the changes below. 

**Major Features**:

1.  Accepting curved boundary supergrids and bathymetry (https://github.com/CROCODILE-CESM/regional-mom6/pull/13)
2. Implement Boundary Tides from OSU's TPXO using GFDL NWA25 Code (https://github.com/CROCODILE-CESM/regional-mom6/pull/12)
3. Import experiment configuration from a light config JSON (https://github.com/CROCODILE-CESM/regional-mom6/pull/12)
4. Major bathymetry fixes for land-sea mask (https://github.com/CROCODILE-CESM/regional-mom6/pull/17)
5. Moving experiment specific changes to MOM_override (All OBC, Tides, and other minor changes) and out of MOM_input
6. New GLORYS downloader (that can account for curved boundaries) (https://github.com/CROCODILE-CESM/regional-mom6/pull/12)
7. Tests: Additional tests (tides. rotation, and regridding), Conftest.py, Add to Docker Image (https://github.com/CROCODILE-CESM/regional-mom6/pull/34)
8. Changing function names to verb beginnings (initial_condition -> setup_initial_condition) (https://github.com/CROCODILE-CESM/regional-mom6/pull/12)
9. Add properties (i.e. more functional object variables) and dataset plotting (https://github.com/CROCODILE-CESM/regional-mom6/pull/12)
10. Add Rotation: Add MOM6 angle calculation at t points (MOM6 doesn't use an input hgrid's angle_dx field. It calculate the angle internally at t-points) & two angle approximation methods for the boundary q/u/v points (https://github.com/CROCODILE-CESM/regional-mom6/pull/34)
11. Add Land Mask (https://github.com/CROCODILE-CESM/regional-mom6/pull/33)

**Minor Features**:

1. Ability to code without greek letters
2. Removed unintentional machine-specific code
3. Increased print statements

**Breaking Changes** (for backwards compatibility):

1. Functions changed names, so it's not compatible with older code. All that needs to be updated is the function name itself in previous code (like the demos). (https://github.com/CROCODILE-CESM/regional-mom6/pull/12)

**Documentation Changes**:

[...]

PR #205, opened 2025-04-25
Rotation Test Fixes + Rotation Calculation Fixes

manishvenu:
**Backstory & Details:**
@navidcy created a test case to test the angle calculation in rotation.py. It's failing! Navid pointed out it's got something to do with the len_lon arg in the calculation, which goes into the modulo_around_point function we took from MOM6.

**Solution/What Happened:**
This is a two part error!

When we do rotation of a unit square in this test, the len_lon changes -> a unit square rotated around the origin stays constant in area but changes in the x range and y range. We can't just put 2 degrees into the function. So, I added a len_lon calculation to the test
This is an edge case, but when the value provided to modulo_around_point is at the very edge of the domain (as in a corner point) the modulo will jump it to the other side of the boundary! We don't want that. So, I added a (len_lon+1e-10) to make sure the edge points stay where they are.

**What is modulo_around_point?**
After looking at it again, modulo_around_point is supposed to return the modulo value of x in an interval [x0-(len_lon/2), x0+(len_lon/2)]. For our unit test example, where we have literally a square, each corner is literally on the edge of the range. The python implementation wasn't inclusive of the boundary, so we slightly bump len_lon to make sure that is. If this is confusing, take an example. We have x = 0.75 degrees, x0 is 0 degrees, len_lon is 1.5 degrees. The range [x0-(len_lon/2), x0+(len_lon/2)] would be [-0.75,0.75]. The modulo_around_point might make x become -0.75 because of how the previous implementation worked. The math would work out to (0.75 - (-0.75) % 1.5 + (-0.75)) = -0.75. We don't want that, points at the boundary should stay where they are. The use case for modulo_around_point is if the function operates as expected, it should map -170 degrees to the equivalent positive value in the [-180, 180] or [0, 360] range or whatever len_lon x0 selection we have.


navidcy:
Thanks @manishvenu! I'll have a look today-tomorrow!

docs/angle_calculation.md
```diff
@@ -19,7 +19,7 @@ Steps 1-5 replicate the angle calculation as done by MOM6. Step 6 is an addition
...
 2. At each ``t``-point on the `hgrid`, we find the four adjacent ``q``-points. We adjust each of these longitudes to be in the range of len_lon around the point itself. ({meth}`rotation.modulo_around_point <regional_mom6.rotation.modulo_around_point>`)
-3. We then find the lon_scale, which is the "trigonometric scaling factor converting changes in longitude to equivalent distances in latitudes". Whatever that actually means is we add the latitude of all four of these points from part 3 and basically average it and convert to radians. We then take the cosine of it. As I understand it, it's a conversion of longitude to equivalent latitude distance.
+3. We then find the lon_scale, which is the "trigonometric scaling factor converting changes in longitude to equivalent distances in latitudes". What that means is we add the latitude of all four of these points from part 3, average it, and convert to radians. We then take the cosine of it. It's a factor we can use to convert a difference in longitude to equivalent latitude difference.
 4. Then we calculate the angle. This is a simple arctan2 so y/x.
```
navidcy, line 23:
```suggestion
4. Then we calculate the angle using `numpy.arctan2` from the ratio of the North-South and East-West distances.
```

docs/angle_calculation.md
```diff
@@ -19,7 +19,7 @@ Steps 1-5 replicate the angle calculation as done by MOM6. Step 6 is an addition
...
 2. At each ``t``-point on the `hgrid`, we find the four adjacent ``q``-points. We adjust each of these longitudes to be in the range of len_lon around the point itself. ({meth}`rotation.modulo_around_point <regional_mom6.rotation.modulo_around_point>`)
-3. We then find the lon_scale, which is the "trigonometric scaling factor converting changes in longitude to equivalent distances in latitudes". Whatever that actually means is we add the latitude of all four of these points from part 3 and basically average it and convert to radians. We then take the cosine of it. As I understand it, it's a conversion of longitude to equivalent latitude distance.
+3. We then find the lon_scale, which is the "trigonometric scaling factor converting changes in longitude to equivalent distances in latitudes". What that means is we add the latitude of all four of these points from part 3, average it, and convert to radians. We then take the cosine of it. It's a factor we can use to convert a difference in longitude to equivalent latitude difference.
```
navidcy, line 22: I don't understand the part 3 of the part **Boundary rotation algorithm** section. Meaning, I don't understand why the described process gives us the angle of rotation of the grid cell.

We are describing here what each line of the code we copied from MOM6 is but we don't give more insight to why this computes the angle I've been trying to understand why this works and I can't.

If we can understand why this computes the angle of the grid cell then let's explain it.
If we can't, then there is no need to document what each line of python code is. Code is sorta self-explanatory so this section of the docs does not add anything.

The fact that I wasn't able to understand it and also that it wasn't computing the correct rotation angle for a grid cell that I was constructing by hand via a rotation was making me skeptical when I was reviewing #193.

So I suggest either of the following:

- We remove the **Boundary rotation algorithm** section
- We ask whoever wrote the algorithm to explain to us and then we explain the algorithm in this section? I'm happy to ask, I just don't know who that is. Could we find from git blame of MOM6 code?

regional_mom6/rotation.py
```diff
@@ -118,6 +118,7 @@ def modulo_around_point(x, x0, L):
     float
         ``x`` shifted by an integer multiple of ``L`` to be closer to ``x0``, i.e., within the interval ``[x0 - L/2, x0 + L/2]``
     """
+    L += 1e-10  # The 1e-10 prevents exact values from swapping from exact boundary to the other boundary (i.e. x0 - L/2 to x0 + L/2) of the interval
```
navidcy, line 121: Thanks for the explanation in the PR comment.

Question: isn't it a better solution instead of adding the ad-hock 1e-10 value in the src code to change the test so that it's not an edge case?

tests/test_rotation.py
```diff
@@ -294,7 +298,7 @@ def test_get_rotation_angle(get_curvilinear_hgrid, get_rectilinear_hgrid):
     assert (angle.values == curved_hgrid.angle_dx).all()
     angle = rot.get_rotation_angle(rotational_method, rect_hgrid, orientation=o)
     assert angle.shape == rect_hgrid.x.shape
-    assert (angle.values == 0).all()
+    assert np.all(np.isclose(angle.values, 0, atol=1e-10))
```
navidcy, line 301: let's define an `atol` constant at the top of this file and use it; this way we can adjust it everywhere easily instead of having to go and change it in several places.

navidcy (changes requested):
thanks @manishvenu! here's some comments/questions/suggestions... let me know what you think!

navidcy:
Does this PR addresses this comments here:

https://github.com/COSIMA/regional-mom6/pull/193#discussion_r1976514559

https://github.com/COSIMA/regional-mom6/pull/193#discussion_r1986197340

Mostly my question about systematic error in angles even when grid cell size becomes very small...

https://github.com/COSIMA/regional-mom6/pull/193#discussion_r1977965482

(These were part of the unresolved issues in #193.) You had mentioned that the MOM6_rotation_method is not perfect. Still, I'd expect that the error in rotation angles decrease the for smaller grid cells.

regional_mom6/rotation.py
manishvenu: Here's a couple options to consider:

1. We could take out this function entirely, I don't think we'll ever pass in an hgrid with angles outside the 0 to 360 range
2. We want to keep the 1e-10 because on the edges of the boundary, we would trigger this edge case issue

tests/test_rotation.py
manishvenu: Sure

manishvenu:
> #193 (comment)

Yeah, all these problems were addressed. You should be able to run the test you created without any issues. It's working in the github actions.

docs/angle_calculation.md
manishvenu: Allright, responding to each part:

> I don't understand the part 3 of the part Boundary rotation algorithm section. Meaning, I don't understand why the described process gives us the angle of rotation of the grid cell.

I totally get that it can be hard to understand, I'll respond with some other sources/explanation that might be easier to understand in a day or so.

> If we can understand why this computes the angle of the grid cell then let's explain it.
> If we can't, then there is no need to document what each line of python code is. Code is sorta self-explanatory so this section of the docs does not add anything.
 I think the documentation here helps me understand the code. I think we should keep it.

> The fact that I wasn't able to understand why the algorithm gives the prescribed angle and also that it wasn't computing the correct rotation angle for a grid cell that I was constructing by hand via a rotation was making me skeptical when I was reviewing https://github.com/COSIMA/regional-mom6/pull/193. I mentioned this a few times, e.g., https://github.com/COSIMA/regional-mom6/pull/193#issuecomment-2649840854

Doesn't this PR fix this? That specific test works in Github Actions, you should be able to run the test you created without issues.

> So I suggest either of the following:
> 
> We remove the Boundary rotation algorithm section
> We ask whoever wrote the algorithm to explain to us and then we explain the algorithm in this section? I'm happy to ask, I just don't know who that is. Could we find from git blame of MOM6 code?

I think we should keep this section, it helps explain the steps of the code. We've also talked about this second question, We can git blame for the MOM6 code, but I'm not sure we need to take that step yet.

navidcy: Hm... 

I don't understand what this part of the code does.

https://github.com/COSIMA/regional-mom6/blob/cdd3c2300c8251f9efee869b9e9bdd0fa795029b/regional_mom6/rotation.py#L171-L177

Why `angle` is given by that expression. The section of the documentation we are discussing supposedly explains it; for me it does not.

(The change of the sign btw adds to the mystery.)

If we don't understand this part, then how are we gonna explain it? If this section of the docs is only to help developers understand what the python code is supposed to be doing then does it belong as a comment in the source code?

navidcy: I'd be very keen to ask somebody to explain to me. I'll try to spend some more time again to try to understand it. :(

navidcy:
> > #193 (comment)
> 
> Yeah, all these problems were addressed. You should be able to run the test you created without any issues. It's working in the github actions.

I'll give it a go!

docs/angle_calculation.md
manishvenu: The negative sign is because this is a clockwise angle calculation, and our rotation calculations wants a counter-clockwise angle

navidcy: Could you have a look at #209?

regional_mom6/rotation.py
navidcy: how about if we drop the `L += 1e10` and instead do:

```python
    if L <= 0:
        return x
    if x == x0 + L/2 # ensure that x0 + L/2 does not flip to x0 - L/2
        return x
    else:
        return ((x - (x0 - L / 2)) % L) + (x0 - L / 2)
```

navidcy: That sounds cleaner since we don't modify the `L` argument. What do you think @manishvenu?

navidcy: btw the comment in the src code currently is wrong; the x0-L/2 point always stays at x0-L/2 (even before the fix!)
the problematic edge point is the x0+L/2

manishvenu: This works for me.

docs/angle_calculation.md
manishvenu: Yup

regional_mom6/rotation.py
navidcy: Cool!

manishvenu: Should be added.

navidcy (approved):

manishvenu:
Are you merging or shall I?

regional_mom6/rotation.py
```diff
@@ -120,6 +120,10 @@ def modulo_around_point(x, x0, L):
     """
     if L <= 0:
         return x
+    elif (
+        x == x0 + L / 2
+    ):  # ensure that boundary point x0 + L/2 does not flip to x0 - L/2
```
navidcy, line 125: I guess this is not vector-friendly?

navidcy: (I see the tests fail and this is the only thing that changed?)

manishvenu: Wow! I'll track it down

tests/test_rotation.py
```diff
@@ -139,10 +140,11 @@ def test_mom6_angle_calculation_method_simple_square_grids(angle):
 
     # Define four point points on a square with side of
     # length 2 and centered at (0, 0)
-    top_left = np.array([-1, +1])
-    top_right = np.array([+1, +1])
-    bottom_left = np.array([-1, -1])
-    bottom_right = np.array([+1, -1])
+    square_size = 0.0001
```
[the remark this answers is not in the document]
manishvenu: Should be resolved

regional_mom6/rotation.py
manishvenu: How's this? I looked at it and the only place it's called expects xr.DataArrays...

manishvenu: Allright since this is passing, I'm approving

navidcy: I can have a look later if you like

manishvenu: Ok, since this is only used in one place, I'm not super worried about it. I'll go ahead and merge.

Issue #198, opened 2025-02-04
Move MOM_parameter functions out of regional_mom6.py


manishvenu:
Currently, three functions exist to adjust MOM_input + MOM_override. They are general utility functions that are better off in another file (maybe utils or a seperate module). This is a suggestion by @angus-g.

Functions:

- change_MOM_parameter
- read_MOM_file_as_dict
- write_MOM_file

PR #206, opened 2025-04-25
Small Scientific Reproducibility, Config + MOM_parameter Refactor 

manishvenu:
~Please see: https://github.com/CROCODILE-CESM/regional-mom6/pull/54~

[reproduced here the first post in https://github.com/CROCODILE-CESM/regional-mom6/pull/54 by @navidcy]

Closes https://github.com/COSIMA/regional-mom6/issues/198

### Changes/Explanation:

1. I was working with Gustavo on running NWA Boundary Conditions with regional-mom6. It seemed like the two things that mattered for reproducing boundary conditions (like GFDLs) was the regridding method (e.g. bilinear, source2d, etc..) and fill method (e.g. horizontal or vertical first, in what direction). I lightly added those arguments as 2 variables to the experiment class, the config files, and the regridding functions. 

    The `regridding_method` variable is a string that defaults to "bilinear", and the `fill_method` is a function that defaults to `regridding.fill_missing_data`.

2. Moved the config out of regional_mom6.py to a new script to take that logic out of the main class.

3. Moved all the MOM parameter editing stuff to new script as well
4. Adjust tests as needed, including splitting tides and parameters.

Goal: The main regional_mom6 script is cleaner and we split some of the complexity of regional_mom6 off into seperate simpler scripts


navidcy:
Thanks @manishvenu for this!
@angus-g would you be down to review this?

navidcy:
Clarification question:

> Goal: The main regional_mom6 script is cleaner and we split some of the complexity of regional_mom6 off into seperate simpler scripts

What do you mean by "regional_mom6 script" and "seperate simpler scripts"? What does "script" refers to? You referring to source code? Parts of notebook? User-generated Python `.py` scripts?

regional_mom6/regional_mom6.py
```diff
@@ -610,6 +501,8 @@ class experiment:
         create_empty (bool): If ``True``, the experiment object is initialized empty. This is used for testing and experienced user manipulation.
         expt_name (str): The name of the experiment (for config file use)
         boundaries (List[str]): List of (rectangular) boundaries to be set. Default is ``["south", "north", "west", "east"]``. The boundaries are set as (list index + 1) in MOM_override in the order of the list, and less than 4 boundaries can be set.
+        regridding_method (str): regridding method to use throughout the entire experiment. Default is ``'bilinear'``
```
angus-g, line 504: Other options are?

regional_mom6/regional_mom6.py
```diff
@@ -610,6 +501,8 @@ class experiment:
...
+        fill_method (Function): The fill function to be used after regridding datasets
```
angus-g, line 505: Takes a dataset, returns a dataset? Defaults to backfilling missing data?

regional_mom6/regional_mom6.py
```diff
@@ -1599,7 +1455,13 @@ def setup_ocean_state_boundaries(
                 boundary condition is not masked.
             rotational_method (Optional[str]): Method to use for rotating the boundary velocities.
                 Default is ``EXPAND_GRID``.
+            regridding_method (Optional[str]): The type of regridding method to use. Defaults to self.regridding_method
+            fill_method (Function): Fill method to use throughout the function. Default is ``rgd.fill_missing_data``
```
angus-g, line 1459: I guess the default is _actually_ `self.fill_method`?

regional_mom6/regional_mom6.py
```diff
@@ -810,7 +713,7 @@ def __init__(
             input_rundir.symlink_to(self.mom_run_dir.resolve())
 
     def __str__(self) -> str:
-        return json.dumps(self.write_config_file(export=False, quiet=True), indent=4)
+        return json.dumps(Config.save_to_json(export=False, quiet=True), indent=4)
```
angus-g, line 716: Do you need to pass `self` here? Also, `Config` doesn't have a `quiet` flag.

```suggestion
        return json.dumps(Config.save_to_json(self, export=False), indent=4)
```

regional_mom6/config.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def save_to_json(obj, path=None, export=True):
+        """Write a ``json`` configuration file for the experiment. The ``json`` file contains
+        the experiment variable information to allow for easy pass off to other users, with a strict computer
+        independence restriction. It also makes information about the expirement readable, and
+        can be also useful for simply printing out information about the experiment.
+
+        Arguments:
+            obj (rm6.Experiment): The experiment object to save.
+            path (str): Path to write the config file to. If not provided, the file is written to the ``mom_run_dir`` directory.
```
angus-g, line 25: If `path=None` (default) and `export=True` (default), it'll try to `open(None, "w")`

regional_mom6/config.py
```diff
@@ -0,0 +1,144 @@
...
+                    "module": value.__module__,
+                }
+            else:
+                config["args"][key] = {
+                    "type": type(value).__name__,
+                    "value": convert_value_to_string(value),
+                }
+
+        if export:
+            with open(path, "w") as f:
```
angus-g, line 45: This is where there's no checking on the value of `path`.

regional_mom6/config.py
```diff
@@ -0,0 +1,144 @@
...
+        filename,
+        mom_input_dir="mom_input/from_config",
+        mom_run_dir="mom_run/from_config",
+        create_hgrid_and_vgrid=True,
+    ):
+        """
+            Load experiment variables from a configuration file and generate the horizontal and vertical grids (``hgrid``/``vgrid``).
+        Computer-specific functionality eliminates the ability to pass file paths.
+
+        (This is basically another way to initialize and experiment.)
```
angus-g, line 60:
```suggestion
        (This is basically another way to initialize an experiment.)
```

regional_mom6/config.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def load_from_json(
+        filename,
+        mom_input_dir="mom_input/from_config",
+        mom_run_dir="mom_run/from_config",
+        create_hgrid_and_vgrid=True,
+    ):
+        """
+            Load experiment variables from a configuration file and generate the horizontal and vertical grids (``hgrid``/``vgrid``).
+        Computer-specific functionality eliminates the ability to pass file paths.
```
angus-g, line 58: Not really sure what this is implying?

regional_mom6/utils.py
```diff
@@ -384,3 +385,23 @@ def is_rectilinear_hgrid(hgrid: xr.Dataset, rtol: float = 1e-3) -> bool:
...
+        for path in paths:
+            all_files.extend(Path(path).glob(pattern))
+
+    if len(all_files) == 0:
+        if error_message is None:
+            return "No files found at the following paths: {} for the following patterns: {}".format(
+                paths, patterns
+            )
+        else:
+            return error_message
```
angus-g, line 406: Seems like this should raise an exception rather than breaking type contract and returning a string?

manishvenu:
> Clarification question:
> 
> > Goal: The main regional_mom6 script is cleaner and we split some of the complexity of regional_mom6 off into seperate simpler scripts
> 
> What do you mean by "regional_mom6 script" and "seperate simpler scripts"? What does "script" refers to? You referring to source code? Parts of notebook? User-generated Python `.py` scripts?

If you look at the diff in the PR, there's two new python files in the package. This code used to be part of regional_mom6.py, and I moved it to seperate files in response to a comment by @angus-g (MOM parameter tools) and my own ideas. If you look at the bullet points in the first comment it expands on this a little bit.

regional_mom6/regional_mom6.py
manishvenu: We do, thanks, I'll write a test case for this.

regional_mom6/config.py
manishvenu: Yup, thanks

regional_mom6/config.py
manishvenu: Thanks!

regional_mom6/regional_mom6.py
manishvenu: Yes, thanks!

regional_mom6/regional_mom6.py
manishvenu: Added more of a comment, thanks

regional_mom6/regional_mom6.py
manishvenu: Done

regional_mom6/config.py
manishvenu: Removing it!

regional_mom6/utils.py
manishvenu: Yup! Thanks

regional_mom6/config.py
manishvenu: Added

navidcy:
@manishvenu there are some conflicts to resolve when you can.

regional_mom6/MOM_parameter_tools.py
```diff
@@ -0,0 +1,203 @@
...
+def change_MOM_parameter(
+    directory, param_name, param_value=None, comment=None, override=True, delete=False
+):
+    """
+    **Requires MOM parameter files present in the run directory**
+
+    Changes a parameter in the ``MOM_input`` or ``MOM_override`` file. Returns original value, if there was one.
+    If `delete` keyword argument is `True` the parameter is removed. But note, that **only** parameters from
+    the ``MOM_override`` are be deleted since deleting parameters from ``MOM_input`` is not safe and can lead to errors.
+    If the parameter does not exist, it will be added to the file.
```
navidcy, line 16:
```suggestion
    Changes a parameter in the ``MOM_input`` or ``MOM_override`` file. Returns original value, if there was one.
    If `delete` keyword argument is `True` the parameter is removed. But note, that **only** parameters from
    the ``MOM_override`` are be deleted, since deleting parameters from ``MOM_input`` is not safe and can lead to errors.
    If the parameter does not exist, it will be added to the file.
    
    **MOM parameter files need to be present in the run directory**
```

navidcy: > "...will be added to the file."

Which file? The`MOM_override`?

manishvenu: Yup, added it

navidcy:
@angus-g could you have another look here? thanks!

regional_mom6/regional_mom6.py
```diff
@@ -14,18 +14,17 @@
...
 import pandas as pd
 from pathlib import Path
-import glob
-from collections import defaultdict
 import json
-import copy
+from regional_mom6 import MOM_parameter_tools as mpt
 from regional_mom6 import regridding as rgd
 from regional_mom6 import rotation as rot
+from regional_mom6.config import Config
```
navidcy, line 21: are these new modules/submodules introduced by this PR?
do we need to include them here?
https://github.com/CROCODILE-CESM/regional-mom6/blob/minimal_scientific_reprod_cherrypick/docs/api.rst?plain=1

manishvenu: Yeah, that's a good point, I'll add them to the autodocs

angus-g (commented):
Seems reasonable, thanks. I'll leave approval to @navidcy

navidcy (approved):

PR #209, opened 2025-05-01
Explain algorithm that computes the angle of every grid cell

navidcy:


navidcy:
Question: the documentation suggests that we compute the angels for all interior t-points + the boundary q-points. How about the interior q-points; don't we compute the angles the correspond to those points?

navidcy:
Actually, reading in detail 
https://github.com/mom-ocean/MOM6/blob/05d8cc395c1c3c04dd04885bf8dd6df50a86b862/src/initialization/MOM_shared_initialization.F90#L572-L587
they compute the diagonal BR[1,0] - TL[0, 1] (which is the minus of the green one we have in the docs now).

So FORTRAN implementation does this:

angle_via_diagonals-old2

And therefore does not need to have a minus after the angle because it's already counter-clockwise and the angle exactly what `angle_dx` expects, that is the deviation of the grid's "East"-"West" orientation from the true East-West.

Should we change our docs and implementation to mimic exactly what FORTRAN code does?

navidcy:
I actually change it to follow mom6
it's even simpler; no need to reverse the angle, etc.

manishvenu:
> Question: the documentation suggests that we compute the angels for all interior t-points + the boundary q-points. How about the interior q-points; don't we compute the angles the correspond to those points?

We calculate the angle at all points in the hgrid, so this documentation could be reworded for all points. Explicitly, we use the angle calculation for u & v points in the initial condition, and ALL points on the boundaries.

docs/angle_calculation.md
```diff
@@ -1,39 +1,37 @@
...
 ## Default Behavior
 regional-mom6 by default computes the the angle of curved horizontal grids (``hgrids``) using the same algorithm as MOM6 does.
-This algorithm is detailed below.
 
-## Detailed Explanation
+We explain below the implementation of MOM6 angle calculation in regional-mom6, which is the process by which regional-mom6 calculates the angle of curved horizontal grids (``hgrids``).
 
-Here we explain the implementation of MOM6 angle calculation in regional-mom6, which is the process by which regional-mom6 calculates the angle of curved horizontal grids (``hgrids``).
+### Grid-cell angle of rotation algorithm
+Steps 1 through 4 replicate the angle calculation in the interior ``t``-points, as done by MOM6 (see {meth}`rotation.mom6_angle_calculation_method <regional_mom6.rotation.mom6_angle_calculation_method>`). Step 5 describes how we can apply this algorithm to the boundary ``q``-points.
```
manishvenu, line 15:
```suggestion
Steps 1 through 4 replicate the angle calculation in the interior ``t``-points, as done by MOM6 (see {meth}`rotation.mom6_angle_calculation_method <regional_mom6.rotation.mom6_angle_calculation_method>`). Step 5 describes how we can apply this algorithm to all boundary points (``q``, ``u``, and ``v`` points).
```

docs/angle_calculation.md
```diff
@@ -1,39 +1,37 @@
...
 
-## Boundary rotation algorithm
-Steps 1-5 replicate the angle calculation as done by MOM6. Step 6 is an additional step required to apply this algorithm to the boundary points.
+1. Determine the longitudinal extent of our domain, or periodic range of longitudes. For grids that go around the globe ``len_lon = 360``; for regional domains ``len_lon`` is provided by the ``hgrid``.
+2. Find the four ``q``-points surrounding each ``t``-point in our grid. These four ``q`` points form a quadrilateral; let's denote them as top-left (TL), top-right (TR), bottom-left (BL), bottom-right (BR). We want to determine the average angle of rotation of this quadrilateral compared to the North-South direction. We ensure that the longitudes of the ``q`` points are within the range of ``len_lon`` around the ``t``-point itself via ({meth}`rotation.modulo_around_point <regional_mom6.rotation.modulo_around_point>`).
```
manishvenu, line 18: Nice, I like this explanation

docs/angle_calculation.md
```diff
@@ -1,39 +1,37 @@
...
+4. To determine the average rotation of the quadrilateral we form the vectors of its two diagonals (the ones that start from the left points and end on the right ones) and then the vectorial sum of the two diagonals. The angle that vector of the sum of the diagonals forms with the North-South direction is a good approximation for the average angle of the quadrilateral and thus this is the angle we associate with the ``t`` point. The diagram below shows two quadrilaterals: the left one is already oriented along longitude-latitude coordinates and therefore the sum of the diagonals is oriented along North-South direction; the one on the right is rotated counter-clockwise by a negative acute angle compared to North-South direction.
+   ![Logo](_static/images/angle_via_diagonals.png)
+   We compute the angle of the sum-of-diagonals vector with the East-West direction via `numpy.arctan2(x, y)`; note that `numpy.arctan2(x, y)` returns `atan(x/y)`. We ensure that we use a counter-clockwise convention for the angle and also convert the angel to degrees.
```
manishvenu, line 22:
```suggestion
   We compute the angle of the sum-of-diagonals vector with the East-West direction via `numpy.arctan2(x, y)`; note that `numpy.arctan2(x, y)` returns `atan(x/y)`. We ensure that we use a counter-clockwise convention for the angle and also convert the angle to degrees.
```

docs/angle_calculation.md
```diff
@@ -1,39 +1,37 @@
...
-4. Then we calculate the angle. This is a simple arctan2 so y/x.
-    1. The "y" component is the addition of the difference between the diagonals in longitude (adjusted by modulo_around_point in step 3) multiplied by the lon_scale, which is our conversion to latitude.
-    2. The "x" component is the same addition of differences in latitude.
-    3. Thus, given the same units, we can call arctan to get the angle in degrees
+   Since the boundaries for a regional MOM6 domain are `q` points and not on the `t` points, to calculate the angle for those boundary points we extend the grid by a bit; see the {meth}`rotation.create_expanded_hgrid <regional_mom6.rotation.create_expanded_hgrid>` method. Doing so, we can apply the method above since now there exist four `t`-points around each boundary `q` point that form a quadrilateral around the `q` points.
```
manishvenu, line 25:
```suggestion
   Since the boundaries for a regional MOM6 domain are `q`, `u`, and `v` points and not on the `t` points, to calculate the angle for those boundary points we extend the grid by a bit; see the {meth}`rotation.create_expanded_hgrid <regional_mom6.rotation.create_expanded_hgrid>` method. Doing so, we can apply the method above since now there exist four points around each boundary point that form a quadrilateral.
```

docs/angle_calculation.md
```diff
@@ -1,39 +1,37 @@
...
 
-5. **Additional step to apply to boundaries**
-Since the boundaries for a regional MOM6 domain are on the `q` points and not on the `t` points, to calculate the angle at the boundary points we need to expand the grid. This is implemented in the {meth}`rotation.create_expanded_hgrid <regional_mom6.rotation.create_expanded_hgrid>` method.
+**Remark**: The angle formed by the curved grid's local direction of constant ``x`` coordinate compared to the North-South and the angle formed by the local direction of constant ``y`` coordinate compared to the East-West are the same. The latter is what is expected in `angle_dx` and this is one reason we ensure we have counter-clockwise convention.
```
manishvenu, line 27: Could you expand upon this? Here's what I got: The angle from x from true north is the same as y from true east.

How does that ensure a counter-clockwise convention? The rest of the comment is great, though not sure why we're highlighting it.

docs/angle_calculation.md
navidcy: thanks!

regional_mom6/rotation.py
```diff
@@ -133,8 +133,8 @@ def mom6_angle_calculation_method(
     point: xr.DataArray,
 ) -> xr.DataArray:
     """
-    Calculate the angle of the grid point using the MOM6 method adapted from the
-    MOM6 code: https://github.com/mom-ocean/MOM6/blob/05d8cc395c1c3c04dd04885bf8dd6df50a86b862/src/initialization/MOM_shared_initialization.F90#L572-L587
+    Calculate the angle of the grid point's local constant-y direction compared to East-West direction
```
manishvenu, line 136: "local constant-y direction" makes sense but is pretty confusing to me. I think it is more common to reference true north & grid north (happy if someone thinks differently). I would change to something like 
```suggestion
    Calculate the angle of the grid point's local constant-x direction compared to the true north direction
```
across the documentation. Especially since that's the way MOM6 does it.

regional_mom6/rotation.py
```diff
@@ -151,7 +151,7 @@ def mom6_angle_calculation_method(
     Returns
     -------
     xr.DataArray
-        The angle that corresponds to the grid point
+        The angle angle of the grid point's local constant-y direction compared to East-West direction.
```
manishvenu, line 154:
```suggestion
        The angle of the grid point's local constant-y direction compared to East-West direction.
```

regional_mom6/rotation.py
```diff
@@ -168,13 +168,26 @@ def mom6_angle_calculation_method(
...
+    TR_BL_diagonal_x = cos_meanlat * (lonB[1, 1] - lonB[0, 0])
+    TR_BL_diagonal_y = top_right.y - bottom_left.y
+
+    # Sum of diagonals components
+    sum_of_diagonals_x = TR_BL_diagonal_x + TL_BR_diagonal_x
+    sum_of_diagonals_y = TR_BL_diagonal_y + TL_BR_diagonal_y
+
+    # Angle of sum-of-diagonals vector with the North-South direction
+    # Note: the minus sign changes convention from clockwise to counter-clockwise
+    angle = -np.arctan2(sum_of_diagonals_x, sum_of_diagonals_y)  # = - atan(x/y)
```
manishvenu, line 187: So we did need the negative? If so, that's great news, because MOM6 says this is a clockwise angle calculations

navidcy:
Is there anything else regarding boundary points I should change @manishvenu?

manishvenu:
> Is there anything else regarding boundary points I should change @manishvenu?

Yup! Sorry forgot to add that comment to the review it should be there now.

regional_mom6/rotation.py
navidcy: The way the code was needed we need it. But if I followed correctly the MOM6 code they compute vectorial sum of two different diagonals and thus their angle is not the one between true North and the pos-y direction of the cell but is rather the one between true East and the pos-x direction of the cell.

regional_mom6/rotation.py
manishvenu: Yeah, that makes sense. That angle is the same as the local x from north-south right?

regional_mom6/rotation.py
navidcy: Actually forget my comment!
What you had was what MOM6 does!

I got confused with the 0-indexing and 1-indexing! My bad.

navidcy:
> > Actually, reading in detail https://github.com/mom-ocean/MOM6/blob/05d8cc395c1c3c04dd04885bf8dd6df50a86b862/src/initialization/MOM_shared_initialization.F90#L572-L587 they compute the diagonal BR[1,0] - TL[0, 1] (which is the minus of the green one we have in the docs (up to this commit).
> > So FORTRAN implementation does this:
> > angle_via_diagonals-old2
> > And therefore does not need to have a minus after the angle because it's already counter-clockwise and the angle exactly what `angle_dx` expects, that is the deviation of the grid's "East"-"West" orientation from the true East-West.
> > Should we change our docs and implementation to mimic exactly what FORTRAN code does?
> 
> I'm not quite following this, but, just to make sure, we reverted back to using the negative sign right? I think that's correct because MOM6 describes this as a clockwise angle calculation. I can relook about it to make sure it make sense if need be!

forget this! it was my confusion ...

This PR didn't change the implementation (I changed it and revert back!).
I only changed some variables to enhance clarity.

regional_mom6/rotation.py
manishvenu: Huh, yeah, sounds good. I think I remember that.

navidcy:
> I had a few minor comments. Let me know what you think. Two questions:
> 
> 1. Are we applying [Rotation Test Fixes + Rotation Calculation Fixes #205](https://github.com/COSIMA/regional-mom6/pull/205) first and then applying this one? If not, we need to add a couple things to this PR.

Let's merge this. And then work on the #205. From the #205 the docs part will be irrelevant but the modulo function part and tests are relevant.

> 2. What did you think about the true north/grid north vs contant-y/east-west terms?

yes, I'll rephrase...
I don't like the term "true", what's the "fake" north?

How about we use "North" and then "cell's local positive y-direction"?

manishvenu:
> cell's local positive y-direction

I'm not gonna argue about that, only mentioned it because the comments in MOM6 say true north. If this makes sense to you , I'm happy to use that. If it's easy, can you add those labels to the diagram?

navidcy:
> > cell's local positive y-direction
> 
> I'm not gonna argue about that, only mentioned it because the comments in MOM6 say true north. If this makes sense to you , I'm happy to use that. If it's easy, can you add those labels to the diagram?

MOM6 mentions "logical North" and "true North".
Does "logical North" reads intuitive to you?

How about we use "local positive `y`-direction" and "North" and we also add a note for the correspondence in MOM6 code comments.

manishvenu:
> > > cell's local positive y-direction
> > 
> > 
> > I'm not gonna argue about that, only mentioned it because the comments in MOM6 say true north. If this makes sense to you , I'm happy to use that. If it's easy, can you add those labels to the diagram?
> 
> MOM6 mentions "logical North" and "true North". Does "logical North" reads intuitive to you?
> 
> How about we use "local positive `y`-direction" and "North" and we also add a note for the correspondence in MOM6 code comments.

Yeah, that sounds perfect.

navidcy:
btw the "logical and true North" mention in MOM6 is only in some few-word-long comments in the source code, right?

manishvenu:
> btw the "logical and true North" mention in MOM6 is only in some few-word-long comments in the source code, right?

Yup, It makes sense to me because I don't think in X/Y very well, hah.

docs/angle_calculation.md
navidcy: I deleted it!

navidcy:
> 1. Refactored part of the angle calculation in terms of diagonals instead of points. I don't personally mind either way. If anyone else reviewing this could give their opinion, I'm happy with it.

No need for everybody to review. If you are happy that's good.

Btw I don't know what you mean "refactor in terms of diagonals instead of points". I didn't refactor anything. It was already written as it's written now. But I just interpreted why that code gave that answer, explained in the docs and just ensured that the code followed the notation in the docs so if someone sees the code after the docs they know what's happening straight away. 

What do you mean by "instead of points"?

The algorithm is using analytical geometry to compute the vector components of the diagonals. I didn't refactor it to do that. I just realised what it was doing. Took me a while because it was a bit cryptic and hidden.

navidcy:
> > btw the "logical and true North" mention in MOM6 is only in some few-word-long comments in the source code, right?
> 
> Yup, It makes sense to me because I don't think in X/Y very well, hah.

How does it read now? Do you think it creates confusion? Shall I add "true" next to North?

manishvenu:
> > > btw the "logical and true North" mention in MOM6 is only in some few-word-long comments in the source code, right?
> > 
> > 
> > Yup, It makes sense to me because I don't think in X/Y very well, hah.
> 
> How does it read now? Do you think it creates confusion? Shall I add "true" next to North?

It's great

docs/angle_calculation.md
```diff
@@ -1,39 +1,35 @@
...
-4. Then we calculate the angle. This is a simple arctan2 so y/x.
-    1. The "y" component is the addition of the difference between the diagonals in longitude (adjusted by modulo_around_point in step 3) multiplied by the lon_scale, which is our conversion to latitude.
-    2. The "x" component is the same addition of differences in latitude.
-    3. Thus, given the same units, we can call arctan to get the angle in degrees
+   Since the boundaries for a regional MOM6 domain are `q` points and not on the `t` points, to calculate the angle for those boundary points we extend the grid by a bit; see the {meth}`rotation.create_expanded_hgrid <regional_mom6.rotation.create_expanded_hgrid>` method. Doing so, we can apply the method above since now there exist four `t`-points around each boundary `q` point that form a quadrilateral around the `q` points.
```
manishvenu, line 25: Boundaries are q, u, and v points, just FYI. 
```suggestion
   Since the boundaries for a regional MOM6 domain are `q`, `u`, and 'v' points and not on the `t` points, to calculate the angle for those boundary points we extend the grid by a bit; see the {meth}`rotation.create_expanded_hgrid <regional_mom6.rotation.create_expanded_hgrid>` method. Doing so, we can apply the method above since now there exist four points around each boundary point that form a quadrilateral.
```

regional_mom6/rotation.py
```diff
@@ -133,8 +133,10 @@ def mom6_angle_calculation_method(
     point: xr.DataArray,
 ) -> xr.DataArray:
     """
-    Calculate the angle of the grid point using the MOM6 method adapted from the
-    MOM6 code: https://github.com/mom-ocean/MOM6/blob/05d8cc395c1c3c04dd04885bf8dd6df50a86b862/src/initialization/MOM_shared_initialization.F90#L572-L587
+    Calculate the angle of the grid point's local x-direction compared to East-West direction
+    using the MOM6 method adapted from: https://github.com/mom-ocean/MOM6/blob/05d8cc395c1c3c04dd04885bf8dd6df50a86b862/src/initialization/MOM_shared_initialization.F90#L572-L587
+
+    Note: this is exactly the same as the angle of the grid point's local y-direction compared to North-Southe direction.
```
manishvenu:
```suggestion
    Note: this is exactly the same as the angle of the grid point's local y-direction compared to North-South direction.
```

manishvenu (approved):
Great

Issue #208, opened 2025-04-28
Package layout breaks tests for conda package


angus-g:
We include the "premade run directories" as a data-only subpackage of the library. However, the file layout in the repository is different to that of the installed package.

In the repository, the `demos` directory is alongside the `regional_mom6` directory:
```
regional-mom6-git
├── demos
│   ├── premade_run_directories
│   │   ├── ...
├── ...
├── regional_mom6
│   ├── ...
│   ├── regional_mom6.py
```

But because we don't want to pollute the Python package namespace with a bare `demos` directory, it is installed as a subpackage:
```
regional-mom6-pkg
├── demos
│   └── premade_run_directories
│       ├── ...
├── ...
├── regional_mom6.py
```

In the `experiment.setup_run_directory()` method, there is the provision for searching both alternatives. However, in the `test_change_MOM_parameter` test, it *only* uses the former arrangement. This causes the conda verification tests to fail, so we can't publish the new version of the package to conda.

```python3
        # Copy over the MOM Files to the dump_files_dir
        base_run_dir = Path(
            os.path.join(
                importlib.resources.files("regional_mom6").parent,
                "demos",
                "premade_run_directories",
            )
        )
       shutil.copytree(base_run_dir / "common_files", expt.mom_run_dir, dirs_exist_ok=True)
```

Resulting in:
```
>       with os.scandir(src) as itr:
E       FileNotFoundError: [Errno 2] No such file or directory: '$PREFIX/lib/python3.9/site-packages/demos/premade_run_directories/common_files'
```

navidcy:
Good catch!

Could you suggest a solution?

PR #213, opened 2025-05-16
Handle both installation possibilities in parameter test

angus-g:
Not a solution to #208 in general, but hopefully makes the tests pass so we can push a new version.


manishvenu (approved):
Looks great to me, considering the explanation in #208. Test runs locally for me. If need be, happy to rework this test, but if this works, that's great!

navidcy (approved):

navidcy:
Do we know if it solve the release issue yet?

navidcy:
Seems so: https://anaconda.org/conda-forge/regional-mom6
woohooo!

PR #214, opened 2025-05-20
Rework OBC Mask

manishvenu:
Details:
I've been taking a look at the OBC mask for our BGC stuff, and think I've come to an improvement/happy solution-> But it definetely needs to be tested on other people's computers!

Right now, our two big OBC generating functions, regrid_velocity_tracers & `regrid_tides`, pass their datasets through a `mask_dataset` function in `regridding.py`. If bathymetry is provided (and it is not by default), the dataset is masked, if bathy isn't provided, all NaNs are filled with zeros to avoid the failures we were getting in https://github.com/CROCODILE-CESM/regional-mom6/issues/25.

However, the problem(s) are:

1. The mask with bathymetry isn't great, for some reason we needed three additional real values at the coast on land & a real value at the corner -  weird land excpetions. 
2. The more important problem (since the previous problem can just be filled) is that it wasn't clear exactly at what point is the land-ocean cutoff. This is because boundaries are on q-u-v points versus tracer points, which is where depth is given.

The fix is that:

1. It seems like xarray/numpy NaNs just aren't working well in the OBCs, so instead if we fill with a fill value we get a much cleaner mask without those weird land_exceptions.  
2. Through some testing, the land-ocean cutoff is the single q-point beyond the MOM6 mask (MOM6 mask can be found in ocean static as "wet"), which makes a certain amount of sense. I can explain this if need be!

Changes:
1. Update `mask_dataset` and `get_boundary_mask` in `regridding.py` according to the description above.
2. Add a utils function for getting the edge of a dataset to make it cleaner.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -2934,10 +2934,8 @@ class segment:
         }
         segment_out = rgd.mask_dataset(
             segment_out,
-            self.hgrid,
             self.bathymetry,
             self.orientation,
-            self.segment_name,
         )
         encoding_dict = rgd.generate_encoding(
             segment_out,
@@ -3192,9 +3190,7 @@ class segment:
             {"lon": f"lon_{self.segment_name}", "lat": f"lat_{self.segment_name}"}
         )
 
-        ds = rgd.mask_dataset(
-            ds, self.hgrid, self.bathymetry, self.orientation, self.segment_name
-        )
+        ds = rgd.mask_dataset(ds, self.bathymetry, self.orientation)
         ## Perform Encoding ##
 
         fname = f"{filename}_{self.segment_name}.nc"
--- a/regional_mom6/regridding.py
+++ b/regional_mom6/regridding.py
@@ -480,6 +480,7 @@ def get_boundary_mask(
     x_dim_name="lonh",
     y_dim_name="lath",
     add_land_exceptions=True,
+    add_even_cooler_land_exceptions = False,
 ) -> np.ndarray:
     """
     Mask out the boundary conditions based on the bathymetry. We don't want to have boundary conditions on land.
@@ -503,6 +504,8 @@ def get_boundary_mask(
         The boundary mask
     """
 
+    if add_land_exceptions and add_even_cooler_land_exceptions:
+        raise ValueError("Can't have both land exceptions true")
     # Hide the bathy as an hgrid so we can take advantage of the coords function to get the boundary points.
 
     # First rename bathy dims to nyp and nxp
@@ -567,12 +570,8 @@ def get_boundary_mask(
                 land  # u/v point on the second level just like mask2DCu
             )
             boundary_mask[(i * 2)] = land
-
-    if add_land_exceptions:
-        # Land points that can't be NaNs: Corners & 3 points at the coast
-
-        # Looks like in the boundary between land and ocean - in NWA for example - we basically need to remove 3 points closest to ocean as a buffer.
-        # Search for intersections
+    
+    if add_even_cooler_land_exceptions or add_land_exceptions:
         coasts_lower_index = []
         coasts_higher_index = []
         for index in range(1, len(boundary_mask) - 1):
@@ -580,6 +579,31 @@ def get_boundary_mask(
                 coasts_lower_index.append(index)
             elif boundary_mask[index + 1] == land and boundary_mask[index] == ocean:
                 coasts_higher_index.append(index)
+        # Corner Q-points defined as land should be zeroed out
+        if boundary_mask[0] == land:
+            boundary_mask[0] = np.nan
+        if boundary_mask[-1] == land:
+            boundary_mask[-1] = np.nan
+
+                
+    if add_even_cooler_land_exceptions:
+        # add the q point on the land side 
+        for i in range(1):
+            for coast in coasts_lower_index:
+                if coast - 1 - i >= 0:
+                    boundary_mask[coast - 1 - i] = ocean
+            for coast in coasts_higher_index:
+                if coast + 1 + i < len(boundary_mask):
+                    boundary_mask[coast + 1 + i] = ocean
+
+
+
+    elif add_land_exceptions:
+        # Land points that can't be NaNs: Corners & 3 points at the coast
+
+        # Looks like in the boundary between land and ocean - in NWA for example - we basically need to remove 3 points closest to ocean as a buffer.
+        # Search for intersections
+        
 
         # Remove 3 land points from the coast, and make them zeroed out real values
         for i in range(3):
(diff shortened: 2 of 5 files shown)


manishvenu:
@alperaltuntas Would love your review on this as well!

helenmacdonald:
Thanks @manishvenu!
Good to hear you have a solution! I would like to check it on a couple of domains here but have a busy couple of days so will review mid next week.

navidcy:
thanks @helenmacdonald!

manishvenu:
Looks like it's not working for tides. But to add it for `velocity_tracers`, or `tides`, just add thebathymetry path to the respective functions. `setup_ocean_state_boundaries `or `setup_tides`

navidcy:
@helenmacdonald are you still happy to review this?

helenmacdonald:
Yes @navidcy - my apologies - a couple of things came up. I am back in the office on Wednesday and will start reviewing then.

helenmacdonald:
Hi @manishvenu , I am looking over this now. Just checking, if there a way for a user on our demo notebooks to set their own fill_value? I don’t think we can set it through the higher level `setup_ocean_state_boundaries` routine. Looks like we can only set the fill_value if we directly call `mask_dataset`?  For most users, I believe that this will then automatically set to a small value.

helenmacdonald:
Thanks @manishvenu  – I think it might be worthwhile setting it up to send a fill value right from the top level commands. However, I don’t think it is as easy as setting a single fill value – user will want to input different fill values for temperature and salinity. Will this be hard to set up?
I am also not too sure what the difference is between a land and ocean fill value is - I think I agree that we only need the ocean – but can you please expand on this?

I was thinking last night – how hard would it be (for in the instance that no fill_value is given) to set the fill value as the nearest neighbour ocean grid cell that has a value we trust? It would mean less a priory knowledge on the area is needed (and we wouldn't need to worry about, say, seasonal signals when setting fill values).

manishvenu:
> Thanks @manishvenu – I think it might be worthwhile setting it up to send a fill value right from the top level commands. However, I don’t think it is as easy as setting a single fill value – user will want to input different fill values for temperature and salinity. Will this be hard to set up?
> 


Hey Helen,

Sounds great. I'll attempt to put a fill value at the top if you agree with the rest of this conversation. It's going to be interesting to do a seperate fill value for temp/salinity. We could do like an average, nearest neighbor, or something like that if it works, or just pass in a dictionary with all the variable fill values.

>  I am also not too sure what the difference is between a land and ocean fill value is - I think I agree that we only need the ocean – but can you please expand on this?

The ocean fill value is an actual fill value for the NaNs, and over land, because NaNs have weird interactions in Fortran, instead replace them with an extreme large/small value, the land fill value, so it's obvious if they're being used in the ocean.\

> I was thinking last night – how hard would it be (for in the instance that no fill_value is given) to set the fill value as the nearest neighbour ocean grid cell that has a value we trust? It would mean less a priory knowledge on the area is needed (and we wouldn't need to worry about, say, seasonal signals when setting fill values).

So, there already is a fill method -> we have the `regridding.fill_missing_data` function, which forward fills vertically & horizontally, so we expect that there aren't any NaNs in the boundary at all. Both boundary condition functions (regrid_vt & regrid_tides) use the fill method. After that step, in the off chance that there are still NaNs, it's filled as a last-ditch attempt in this function. If not, it might fail in the run because of those NaNs.

For a more elaborate fill method, I think we need to set some time to figure that out. For the BGC OBCs for MARBL, initially, we've been doing nearest neighbors in the horizontal & then vertical forward filling down. I assume y'all are doing something to, so there's a lot we could/should hash out and write up for this package. 

The other thing is that the fill method can be swapped out as a function by being added in the constructor if it's the same structure as regridding.fill_missing_data. This was part of my thinking to figure out a better fill method for the package.

I think we should make a PR for the fill method to figure this out (or maybe when I come in september)! For now, the `fillna` in this function is just as a backup of the filling method earlier in the OBC workflow. 

Given that, do you think we should have a fill value passed into this last ditch attempt from the top level? Or instead, add it internally in the regrid_* functions. Or maybe do nothing at all!

manishvenu:
> I think I am following now - agreed, we don't need to pipe the fill values to the higher levels. My understanding is that essentially, the fill value is a catch-all in case issues arise, and we have it set to zero. Issues used to arise but with your fixes, we hoping the fill value is not used (but we are not confident on this and would like to investigate the fill method more?). I see you have a useful warning for if the fill value is used.
> My concern is that setting temp and salt to zero is not ideal - we could end up with some suboptimal boundary conditions. How about we add a suggestion to the warning (which is given when the fill value is used) that the user should check the resultant outputs and consider some post processing to manually replace oddly small values with nearby good values (and maybe ask users to report that they got this message to help us improve the fill method).

Yeah, that's exactly it. I think we need to investigate the fill method, and I can add a warning when this catch-all is used.

How's this:
We can just throw an warning (should it be an error?) to say you have missing values over ocean (if bathymetry is provided), if bathymetry isn't provided, a warning to say there are missing values, and we don't know whether they are land/ocean. We could provide how we filled it (and we have to fill it or MOM6 will throw an error) and how the user could adjust it as well (keep encodings, replace the abnormal values, save with encodings). 

In the case where bathymetry isn't provided, we could then fill the NaNs to extremely large values to avoid the issues with NaNs in fortran. 

> I don’t think anyone at this end is doing any preprocessing of data to fill land cells (the BGC being developed uses the native tools from this toolbox) so I think a discussion in September about the fill method would be useful.

Sounds great! I'm using the toolbox to, but did filling on the source grid upon advice from others in the lab. Talk in September sounds great!

>I will do some tests and finish up the review early next week.

Thanks!

helenmacdonald:
Hi @manishvenu  – I agree with this approach (and a warning rather than an error is sufficient).

manishvenu:
> Thanks!
> 
> If you get a chance before I do, the bit of code that removed `ntiles` was in `regrinding.py` (old lines 528):
> 
> ```
> 
>    # Identify any extra dimension like ntiles
> 
>     extra_dim = None
> 
>     for dim in bathy.dims:
> 
>         if dim not in ["nyp", "nxp"]:
> 
>             extra_dim = dim
> 
>             break
> 
>     # Select the first index along the extra dimension if it exists
> 
>     if extra_dim:
> 
>         bathy = bathy.isel({extra_dim: 0})
> 
> ```
> 
> I was going to suggest re-adding this but it may work better if we can avoid creating the ntiles dimension to start with. 
> 
> 
Sounds great, thanks! I'll let ya know if/when I get to it. If it helps, the CROCODILE topo doesn't have ntiles so I wonder if it's like a FRE-tools thing or smthng .

helenmacdonald (approved):

helenmacdonald:
> I think I've answered all comments, and I believe it is working. Got weird conflicting index issues originally, but they should be fixed!

Thanks - looks good to me - I have approved

manishvenu:
Great Thanks!

PR #220, opened 2025-07-09
Simplify the create_regridder function 

manishvenu:
The `create_regridder` function default looks for the reuse_weights file if it exists using an if statement, instead we can follow this: https://github.com/pangeo-data/xESMF/pull/234#issuecomment-1409401017, which uses os.path.isfile instead.

Fixes https://github.com/CROCODILE-CESM/regional-mom6/issues/58


helenmacdonald (approved):
Thanks - looks good to me

PR #229, opened 2025-08-12
Fre fix

kshedstrom:
run_FRE_tools failed for me with the double paths.


manishvenu:
> Thanks @kshedstrom for adding the notebook – and for raising the FRE tools issue. It has been quite a while since fre-nctools changed the file structure, so I think we really need to apply this fix! I just want to check about backward compatibility first.
> 
> @manishvenu , @ashjbarnes and @navidcy – what are your thoughts on this? Fre-nctools changed their paths a long time ago now. Currently this toolbox has a hardwired path according to the old version of fre-nctools and anyone using a newer version gets errors - so these changes are needed. When we push these changes through there will be some people who were using an old version of fre-nctools will start to get errors. Do we want to have backward compatibility – or encourage everyone to update their fre-nctools?
> 
> We could do backward compatibility by forcing everyone to separately specifying the path for each of check_mask, make_solo_mask and make_quick_mosaic in the demo notebooks

I like the idea of defaulting to the new paths, but having arguments to specify paths for each tool as an option for those with older FRE-nctools!

ashjbarnes:
Thanks @kshedstrom ! Yes I agree, let's change to work with new paths for FRE tool. Could we split the new demo notebook into a separate PR? As it is this notebook still needs tidying up (e.g. the text is all still the same as if it was a Tasmania example). It would be really cool to have a final version of this domain included in the demos though down the track, but for now we should definitely just approve Kate's suggestions

PR #237, opened 2025-09-09
add time_units kwarg to wrapper functions for segment

ashjbarnes:
The `Segment` class has a `time_units` keyword argument to specify whether your boundary forcing data is monthly, daily etc. However, the wrapper functions `setup_ocean_state_boundaries` and `setup_single_boundary` didn't include this kwarg. This PR adds the kwarg so that users can specify boundary time units in the demo notebooks

See https://forum.access-hive.org.au/t/issues-with-regional-mom6-when-using-1-degree-resolution-ocean-forcing-data/4565/8?u=ashjbarnes

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1434,6 +1434,7 @@ class experiment:
         varnames,
         arakawa_grid="A",
         bathymetry_path=None,
+        time_units = "days",
         rotational_method=rot.RotationMethod.EXPAND_GRID,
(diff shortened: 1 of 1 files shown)


ashjbarnes:
If this works then @elizabeth-ellison should be able to do monthly forcing. Testing on the existing run where Lizzie hardcoded things to work could be a good test!

ashjbarnes:
Since we've merged #250 , this makes this PR obselete.

PR #240, opened 2025-09-10
Re add ryf

### LINKED ISSUE #239: Where did the repeat_year_forcing go?
The repeat_year_forcing argument isn't used but used to be! Where did we drop it -> should be in the versions before v1!

ashjbarnes:
closes #239


manishvenu (approved):
ryf looks great! Looks like the time_units stuff got mixed in here as well EDIT: time_units looks like won't work here, but this is great.

ashjbarnes:
oh dw it's letting me merge now!

PR #244, opened 2025-09-24
fixing setup_ocean_state_boundaries to work with arakawa_grid="C"

elizabeth-ellison:
Setup_ocean_state_boundaries was not working with arakawa_grid="C" 
Made a small change to l2815 to the regridder u/v velocity function.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1434,6 +1434,7 @@ class experiment:
         varnames,
         arakawa_grid="A",
         bathymetry_path=None,
+        time_units = "days",
         rotational_method=rot.RotationMethod.EXPAND_GRID,
         regridding_method=None,
         fill_method=None,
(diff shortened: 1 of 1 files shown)


ashjbarnes:
Nice! this PR contains the monthly forcing fix too. Did this work @elizabeth-ellison ? Or we still need to test that? If that's all g then I'm happy to merge your C grid fix along with my monthly forcing commits

elizabeth-ellison:
So I tested the monthly forcing fix, and while it works in the rmom6 notebook in the sense the files are generated, those files generated don't actually work when you try and run a model using them.   @dougiesquire suggested "months since" is not date format that is likely to work, and maybe it still needs to be in days since?

helenmacdonald:
> Looks great! Thank you so much! Could you remove the time units changes from this PR so that it's only the "fixing setup_ocean_state_boundaries to work with arakawa_grid="C"" line?

I have cleaned things up a bit so we only get @elizabeth-ellison's bug fix.
[This](https://github.com/COSIMA/regional-mom6/tree/fix_bry_arakawaC) branch just has the arkawa-c bug fix. @elizabeth-ellison can you please create a pull request + merge from there.

[This](https://github.com/COSIMA/regional-mom6/pull/237/commits) pull request also has @ashjbarnes bug fix so we should close one of these.

Issue #249, opened 2025-10-27
Integrating MARBL BGC Boundary Condition workflow from CrocoDash  and WOMBAT BGC workflow into regional-mom6


manishvenu:
**Details:**
Both CrocoDash and regional-mom6 users are starting to use coupled BGC models (MARBL and WOMBAT respectively). It looks like WOMBAT generation is using expt.setup_boundaries, which creates segment objects and then calls regrid_velocity_tracers, which calls regridding module functions. For CrocoDash, I went a little differently and used the direct regridding module functions for MARBL, and given the success using setup_boundaries, I'd like to remerge the MARBL stuff into regional-mom6. 

We have briefly discussed doing this at the NRI workshop and the CROCODILE one, so hopefully this isn't too much of a surprise. 

Below, I thought I'd write down how I'd like to change/add-to regional-mom6 to accomodate BGC tracers (and am more than willing to help code up WOMBAT changes).

- [x] Clean up the segment class constructor. We use the segment class in case we want to take advantage of [this feature](https://cloud.githubusercontent.com/assets/5859571/15801848/bfd69106-2a6f-11e6-9d52-c49133165a78.png). However, it also takes a bunch of GLORYS arguments that should be in velocity tracers. This will make it cleaner to create new segments and keep all of the specific product stuff for the specific function (https://github.com/COSIMA/regional-mom6/pull/250)
- [x] Touch up regrid_velocity_tracers with more arguments and customization (maybe break up the velocity tracers into two seperate functions?)  (https://github.com/COSIMA/regional-mom6/pull/250)
- [x] Add the changes for MARBL BGC (#254)
- [ ] Add the changes needed to make the OBCs for WOMBAT (iirc, things like merging all of the files, change dates, etc..)
- [ ] MARBL BGC uses a fill method that should come from the mom6_bathy PR, so making sure that integrates well with regional_mom6. (Can be a later change, it's no longer critical)

FYI, the MARBL module is here (CROCODILE's version of expt.setup_ocean_state_boundaries): https://github.com/CROCODILE-CESM/CrocoDash/tree/main/CrocoDash/extract_forcings

PR #250, opened 2025-10-28
Clean up Segment Class

manishvenu:
Details:
1. Move the arakawa grid functionality into its own function, with its own validation.
2. Reprocess the variable mapping into a format that allows for curvilinear grids, multiple z dimensions, and clearer variable names, which is needed for MARBL & I think clearer overall!

Changes:

1. Move the product specific items (infile, arakawa grid) to the function where the product is used
3. Create a new function to handle arakawa grid shenanigans (moving it out of regrid_velocity_tracers)
4. Validate the varnames dictionary/variable a lot more and reprocess them (I find it easier to understand this way)
5. Slightly change the regridding process renaming for curvilinear grids.
6. Some time units updates (if time units include the word since, we assume they are already in the correct format and don't need to be rewrote.)

I think this really cleans up regrid_velocity_tracers! It really reads like one smooth workflow now! Regrid, Rotate, Fill, & Format

This is the first task of: https://github.com/COSIMA/regional-mom6/issues/249

### DIFF
--- a/regional_mom6/regridding.py
+++ b/regional_mom6/regridding.py
@@ -317,6 +317,7 @@ def add_or_update_time_dim(ds: xr.Dataset, times, z_dims=None) -> xr.Dataset:
 
     if "time" in ds.dims:
         regridding_logger.debug("Time already in dataset, overwriting with new values")
+        times.attrs = ds.time.attrs
         ds["time"] = times
     else:
         regridding_logger.debug("Time not in dataset, xr.Broadcasting time dimension")


regional_mom6/regional_mom6.py
```diff
@@ -2693,161 +2619,93 @@ def __init__(
...
+            reprocessed_var_map,
+            rawseg,
+            coords,
+            Path(self.outfolder),
+            regridding_method,
+            self.orientation,
+        )
 
-            rotated_ds = xr.Dataset(
+        # TODO: Check if regridding 3 times is noticeably slower on A grids (shouldn't be, the regridders are created efficiently in the first place)
```
helenmacdonald: > # TODO: Check if regridding 3 times is noticeably slower on A grids (shouldn't be, the regridders are created efficiently in the first place)

was this done?

ashjbarnes:
Looks great! I'm using regional-mom6 tomorrow for something new - I'll switch to this branch and use it as an opportunity to test that everything is working as we expect to get a running model on our computer. If that works then let's approve and merge :)

Sorry - I appear to have caused a merge conflict by merging one of my hackathon PRs from last year. Will fix when I review!

helenmacdonald:
Thanks @manishvenu  - I'm happy for you to merge.

manishvenu:
Oh hmm maybe something broke in the merge - I'll plan on generating some files tmrw. One thing to check if you have a second is the time attr on all the files? Maybe it didn't get carried properly somewhere🤷‍♂️. I'll check tmrw tho!

manishvenu:
I think we need it to! Give me a few hours -> will fix it today!

manishvenu:
Ok, so my generated files look like they get the correct time attrs with this setup. Maybe it's an xarray version issue. What are you working with? This is mine
```  Name         Version    Build   Channel
───────────────────────────────────────────
  cf_xarray    0.10.4     pypi_0  pypi   
  pint-xarray  0.6.0      pypi_0  pypi   
  xarray       2023.12.0  pypi_0  pypi 

```

I know there were versions of xarray that did not carry the attrs from these forcing files when you open and then close them again, so maybe that's this issue. 

What we could do is explictly carry over the time attrs into the encoding dict. It would be a repeat, but it wouldn't harm anything and might work.


Something like changing this call:

```
 encoding_dict = {
            "time": {"dtype": "double"},
            f"nx_{self.segment_name}": {
                "dtype": "int32",
            },
            f"ny_{self.segment_name}": {
                "dtype": "int32",
            },
        }
```

to 


```
 encoding_dict = {
            "time": {"dtype": "double"
           "calendar": segment_out.time.attrs["calendar"],
                "units": segment_out.time.attrs["units"],

},
            f"nx_{self.segment_name}": {
                "dtype": "int32",
            },
            f"ny_{self.segment_name}": {
                "dtype": "int32",
            },
        }
```

manishvenu:
Want me to push this attempted change to test it? Or do you want to add it to your checkout first? @ashjbarnes

ashjbarnes:
> ```
>  encoding_dict = {
>             "time": {"dtype": "double"
>            "calendar": segment_out.time.attrs["calendar"],
>                 "units": segment_out.time.attrs["units"],
> 
> },
>             f"nx_{self.segment_name}": {
>                 "dtype": "int32",
>             },
>             f"ny_{self.segment_name}": {
>                 "dtype": "int32",
>             },
>         }
> ```

This fix doesn't work, because ` segment_out.time.attrs` is empty after applying the `rgd.add_or_update_time_dim`. Running on our version of xarray (25.11), and adding print statements to lines 2764-2767 such that

```
print(segment_out.time.attrs)
segment_out = rgd.add_or_update_time_dim(
    segment_out, times, reprocessed_var_map["depth_coord"]
)
print(segment_out.time.attrs)
```

I get a dictionary of attributes, then an empty dictionary. It seems that this method is the culprit, as it doesn't carry on attributes as expected on xarray 25.11

manishvenu:
I'm not sure but probably! All I did was build the regional_mom6 environment, so nothing crazy. (It is at xarray 2025.12)

manishvenu:
Will this work? I just pushed a rearrangement? So maybe that'll work

ashjbarnes:
> Sounds good to me! Please feel free to approve/merge whenever ya like!

ok so the tests fail, confirming that the regridding functions _do_ strip the time data on whatever version of xarray github actions are running for us

@manishvenu 's fix to the segment class means that the time units are re-added afterwards, so this PR is ready to merge, but we do now know that some of the regridding functions aren't behaving as intended if we want them to preserve metadata

I'll revert my testing PR and make an issue about it to remind us that we need to be careful with the regridding functions until they carry over metadata

PR #254, opened 2025-11-11
Pint Addition

manishvenu:
Changes:
In the initial condition and OBC setup, we add a soft pint conversion for u, v, temp, eta, and depth coordinate. (it it fails, it assumes we are in the correct unit). 

1. pint_xarray added to requirements
2. a unit definition file for non-standard units
3. a helper function to attempt unit conversion
4. added the function into setup_initial_condition and regrid_velocity_tracers

Reasoning:
CESM output for MARBL needs this, Dougie recommended using pint.

Questions:

1. There is a conversion from kelvin to celsius in the code, I don't know what that's from, so I haven't touched it.

Note for Merge:
This is set to merge into a branch on COSIMA that is not active (this is so it shows up on the COSIMA PR page)! It should be merge into CROCODILE-CESM/regional-mom6:clean_segment_class

### DIFF
--- a/regional_mom6/regridding.py
+++ b/regional_mom6/regridding.py
@@ -317,6 +317,7 @@ def add_or_update_time_dim(ds: xr.Dataset, times, z_dims=None) -> xr.Dataset:
 
     if "time" in ds.dims:
         regridding_logger.debug("Time already in dataset, overwriting with new values")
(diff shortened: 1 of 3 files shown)


manishvenu:
Hey Helen! Could you add anybody you think should also review this on COSIMA side?

PR #261, opened 2026-01-14
Pint Addition (reopened with respect to main)

manishvenu:
Changes:
In the initial condition and OBC setup, we add a soft pint conversion for u, v, temp, eta, and depth coordinate. (it it fails, it assumes we are in the correct unit).

- pint_xarray added to requirements
- a unit definition file for non-standard units
- a helper function to attempt unit conversion
- added the function into setup_initial_condition and regrid_velocity_tracers

There is warnings/loggings that are printed in association with this.

Reasoning:
CESM output for MARBL needs this, Dougie recommended using pint.

Questions:

There is a conversion from kelvin to celsius in the code, I don't know what that's from, so I haven't touched it.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1078,6 +1078,8 @@ class experiment:
 
         ## if min(temperature) > 100 then assume that units must be degrees K
         ## (otherwise we can't be on Earth) and convert to degrees C
+        ## Although we now attempt a pint convert, we're leaving this manual conversion in for now
+        ## just in case, as K->C is absolutely necessary, and for some inputs pint may fail where this won't.
         if np.nanmin(ic_raw[reprocessed_var_map["tracer_var_names"]["temp"]]) > 100:
             ic_raw[reprocessed_var_map["tracer_var_names"]["temp"]] -= 273.15
             ic_raw[reprocessed_var_map["tracer_var_names"]["temp"]].attrs[


ashjbarnes:
I've tried this branch, and when I delete the manual conversion from C->K already in rmom6, the pint conversion logic doesn't convert the temperature initial condition, even though its units are "K". I do get the warning `could not use pint for data array temp` so it's trying! It also throws the same warning for `u` and `v`, which have values `m/sec`. These are the variable names from ACCESS-OM2

I don't know enough about pint, but I'd assume that "K" would be a pretty obvious unit name that it'd want to convert! Perhaps something not quite right?

manishvenu:
Can you send a sample file? Pint's really finicky (which is why I'm not a huuge fan) and often the files need to be prepped for pint first.

helenmacdonald:
ACCESS-OM3 appears to have `degC`
```
ncdump -h /g/data/ol01/outputs/access-om3-25km/MC_25km_jra_iaf-1.0-beta-5165c0f8/output000/access-om3.mom6.2d.tos.1mon.min.1958.nc

netcdf access-om3.mom6.2d.tos.1mon.min.1958 {
dimensions:
	xh = 1440 ;
	yh = 1152 ;
	time = UNLIMITED ; // (12 currently)
	nv = 2 ;
variables:
	double xh(xh) ;
		xh:units = "degrees_east" ;
		xh:long_name = "h point nominal longitude" ;
		xh:axis = "X" ;
	double yh(yh) ;
		yh:units = "degrees_north" ;
		yh:long_name = "h point nominal latitude" ;
		yh:axis = "Y" ;
	double time(time) ;
		time:units = "days since 1900-01-01 00:00:00" ;
		time:long_name = "time" ;
		time:axis = "T" ;
		time:calendar_type = "GREGORIAN" ;
		time:calendar = "gregorian" ;
		time:bounds = "time_bnds" ;
	double nv(nv) ;
		nv:long_name = "vertex number" ;
	float tos_min(time, yh, xh) ;
		tos_min:_FillValue = 1.e+20f ;
		tos_min:missing_value = 1.e+20f ;
		tos_min:units = "degC" ;
		tos_min:long_name = "Sea Surface Temperature" ;
		tos_min:cell_methods = "area:mean yh:mean xh:mean time: min" ;
		tos_min:cell_measures = "area: areacello" ;
		tos_min:time_avg_info = "average_T1,average_T2,average_DT" ;
		tos_min:standard_name = "sea_surface_temperature" ;
```
So I think we are good in the long-term

ashjbarnes:
@navidcy help!! black formatting is being really weird. I've run the latest black formatter from the gadi analysis environment and pushed these changes. It thinks the files are perfect. But the black stable checker on github actions doesn't like them ?! Could this be a version thing? Perhaps you've seen it before

manishvenu:
Ok, so they fully updated black in the CI?

PR #263, opened 2026-01-20
Add segment axis

ashjbarnes:
This is a tiny quality of life fix that adds X and Y axes to stop the annoying warnings from MOM since it can't certify the axis coordinates for boundary files. 

It's built on top of the pint addition PR. Sorry @manishvenu for pushing to croc - I had my remote here already from testing pa_2 hope that's ok!

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -2888,8 +2888,12 @@ class segment:
         if self.repeat_year_forcing:
             segment_out["time"] = segment_out["time"].assign_attrs({"modulo": " "})
         # Add axes to suppress annoying runtime warnings
-        segment_out[f"ny_{self.segment_name}"] = segment_out[f"ny_{self.segment_name}"].assign_attrs({"axis":"Y"})
-        segment_out[f"nx_{self.segment_name}"] = segment_out[f"nx_{self.segment_name}"].assign_attrs({"axis":"X"})
+        segment_out[f"ny_{self.segment_name}"] = segment_out[
+            f"ny_{self.segment_name}"
+        ].assign_attrs({"axis": "Y"})
+        segment_out[f"nx_{self.segment_name}"] = segment_out[
+            f"nx_{self.segment_name}"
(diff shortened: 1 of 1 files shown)


ashjbarnes:
ahh too bad. I'd hope it would be an easy fix. Failing due to conflicting indices?

manishvenu:
> > No worries on the croc fork! Want me to track down this warning in MOM6? It'll help with the validate files module we gotta add.
> 
> 
> 
> It's this warning
> 
> 
> 
> `categorize_axes: Failed to identify x- and y- axes in the axis list (nx_segment_002, ny_segment_002, time) of a variable being read from ./INPUT/forcing_obc_segment002`
> 
> 
> 
> I can confirm that this micro fix does keep categorize_axes happy, but it's tripping up our testing I guess! Would be nice to resolve since it prints out for _every PE_. You must get it too in your mom6 output right?

Oh! It's only our testing? I'm happy to fix that if ya like.

And yeah, we do get that 2, I think there's an issue somewhere in the list about it.

manishvenu:
1. I think the problem is we already have those variables added.
2. I think tests pass from this commit, if we just assign the attribute? https://github.com/CROCODILE-CESM/regional-mom6/commit/ee520c86dea41421f46f99db20f51e05df16399b. Would this work for MOM6? Ignore the black formatting.
3. Were you able to generate files from this PR? Seems like it should fail. 
4. This specific test is very old I think (and sometimes a pain to debug because its in the docker container)

manishvenu:
There's one test in CrocoDash that fails on the OBC generation which may be another one of those xarray things, which is why I was surprised the PR works (maybe the newer xarray works with it). The only change in the commit was assigning the attribute instead of reassigning the variable.

ashjbarnes:
OHH ok I see why you opened the issue (that I said you can close oops). It'll be a pain if we both use different versions of black in general.

Ok so since your changes avoid the testing headaches, I'll cherry pick your commit, black re-format, then push back to here

PR #238, opened 2025-09-09
Integration of mom6_bathy & regional_mom6

manishvenu:
Co-authored-by: @ashjbarnes 

**Summary**:

Integrate mom6_bathy into regional_mom6. There are NO front-end changes to this PR (the one exception is slight changes to what is printed out in the setup_bathymetry function). 

Shift the grid functions from regional_mom6 to mom6_bathy to *eventually* take advantage of the functions in mom6_bathy. For now, this PR just reproduces the user facing functions. 

This PR will add the mom6_bathy as a dependency (mom6_bathy @ git+https://github.com/NCAR/mom6_bathy.git@master) and won't use the submodule that currently exists in the code folder

**Changes:**
expt.make_hgrid, make_vgrid, setup_bathymetry now use mom6_bathy grid, topo, and vgrid

**Advantages**:

There are a ton of advantages, but some highlights:

**Gain access to useful functions in mom6_bathy:**
1. Runoff Mapping
2. Chlorophyll File Generation for regional domains
3. Generate ESMF Mesh files (for NOUPC Coupling) and CICE grids with ease from a bathmetry file
5. Use the Topo Editor (an interactive widget for bathymetry editing)
6. EDIT: My favorite advantage is using the mom6_bathy fill method! [Here](https://i.imgur.com/NspviXV.png) is an example of how nice this fill method is when you use a low resolution product (like global CESM output).

**Regional_mom6/CROCODILE Advantages**:
7. Merge the grid generation advances CROCODILE & regional_mom6 make together 
8. mom6_bathy gains additional functions in generating hgrids and topos
9. More easily understand what a grid, topo, and vgrid are inside of regional_mom6-> they're objects instead of netcdf files
10. This will go a huge way to making these files more self-describing!

**Disadvantages/Considerations:**
1. Add another layer of complexity to regional_mom6
2. Currently, this development is on a branch of NCAR/mom6_bathy, which can make editing challenging. (I'd totally recommend COSIMA make a fork)
3. The big kicker is that this moves all of the grid generation out of regional_mom6 into mom6_bathy (Since mom6_bathy will be installed as a pip dependency, edits to grid generation would be PRs into the mom6_bathy repo)

Companion PR: https://github.com/NCAR/mom6_bathy/pull/34

**For reviewers/maintainers:**

[...]

PR #266, opened 2026-01-23
Rom3 setup

ashjbarnes:
Overhaul the `setup_run_directory` function so that it works for an ACCESS-rOM3 (NUOPC+payu) configuration, as well as retaining existing functionality (e.g. generic FMS based run, or FMS + payu). 

Now that `mom6_forge` has been included as a dependency of regional-mom6, we're ready to convert this PR from draft. Most of the code is ready, but we are still pending three things prior to merging this PR and deploying rom3 properly:

- [ ] 1. Overhaul the documentation and readme to explain the rOM3 / NUOPC / FMS use cases of regional-mom6. Point users to @manishvenu 's code if they use CIME / CESM frameworks, and explain that they can use the FMS version of regional-mom6 if they want a totally machine agnostic use case
- [ ] 2. ERA5 wind forcing needs to be supported on Gadi with payu for nuopc runs of MOM6. Until we have this, our demos will make no sense as we're forced to use JRA with GLORYS, giving a nonsensical mismatch in surface conditions
- [ ] 3. We need to test the deployment of a regional-mom6 environment on gadi using @atteggiani 's workflow. This environment also needs to work well with notebooks

Main features / changes

- [x] 1. Tidy up the code that used to be in `setup_run_dir`. Use proper parsers like `ruaml` for YAML files to more nicely modify config.yaml, and retain comments for readability
- [x] 2. Split all of the generic setup logic that's needed for _all_ mom6 runs to its own function `setup_generic`. 
- [x] 3. Add `setup_rom3` function which calls `setup_generic` and then performs the remaining steps needed for a rOM3 run
- [x] 4.  Add `setup_FMS` function which continues to generate exactly the same run configuration files as the old `setup_run_directory` function before the refactor. 
- [ ] 5. Add some rudimentary changes to the nuopc config file for the setting of cpus. e.g., user can request `ncpus` as a kwarg to `setup_rom3`, and a default of 50% of CPUs applied to the data atmosphere. This will allow users to more easily scale up / down their domain
- [x] 6. Removed feature: I got rid of the support for the automated handling of FRE-NCtools mask tables. This was a lot of annoying fiddly code that doesn't really help many people now that we're shifting away from FRE-NCtools, and probably to the auto mask table  
- [ ] 7. Integrate with @helenmacdonald 's regional mom6 [template](https://github.com/ACCESS-NRI/access-om3-configs/tree/dev-regional_jra55do_iaf). We haven't figured out how to do this exactly just yet, so currently the `setup_rOM3` function copies my input files. Related to this step: Before deployment we need the template to have ERA5 surface forcing in its `drof_in` file, but this is a work in progress

Minor quality of life changes
- [x] 2. Put timestep changes at the top of the override file to prompt users that they should play with these settings

UPDATE July 2026
This PR has been split, with unchecked features to be added in a future PR. This way the main branch will have the features of rom3, so it's easier to keep rmom6 up to date with NCAR while we're transitioning rom3 rather than FMS as the main way of using rmom6 in Australia. Now this PR only has changes to the setup_* functions and notebooks, none of which will clash with the main code development and don't affect NCAR collaborators


ashjbarnes:
This merge carried with it some fixed I needed to make on this branch. The way that the previous PR #238 which merged the mom6 forge functionality into rmom6 had some issues. The grid and topo properties didn't have setters, which meant that in some scenarios, running `self.topo = Topo(` broke. I've removed instances where `self.topo` is set like this, so that the `topo` and `grid` properties are now only tied to the bathymetry and hgrid files stored on disk. 

I wonder though if constantly reading them from disk as a property will have performance issues for large experiments, vs storing the grids / bathy in memory? @manishvenu  might know? We can revisit this in a future PR!

ashjbarnes:
@helenmacdonald We still need to wait on the ERA5 forcing to actually release this, since the GLORYS boundaries won't match the jra forcing. In fact, I ran the tassie  demo for a month and the boundaries look pretty crap compared to the old FMS glorys / era5 tassie domain, probably due to this mismatch

But in the meantime this notebook is ready for some testing, and the PR can start being reviewed

I still need to revisit the machine agnostic demo and update the documentation there, but what do you think of the documentation I've included in the `rOM3 demo` notebook?

helenmacdonald:
Thanks @ashjbarnes good work! I will take a closer look over the next few days. 

> I ran the tassie demo for a month and the boundaries look pretty crap compared to the old FMS glorys / era5 tassie domain, probably due to this mismatch

I am a bit worried about this - I would expected the mismatch to change the answer, but not give a huge degradation in the model solution (unless there is a quality / resolution issue between the 2 forcing products).  When we have the ERA5 winds we should double check that it fixes things.  In the meantime, I will have a closer look at the physics activated in the stencil as some of this may be different from the original stencil we were using in FMS (unless you have already ruled this out as the cause of the degradation?). 

Are we able to merge into main before we
>  test the deployment of a regional-mom6 environment on gadi using @atteggiani 's workflow

?
I think we may need to have this code in main before we can test it on the workflow.

ashjbarnes:
> Thanks @ashjbarnes good work! I will take a closer look over the next few days.
> 

> I am a bit worried about this - I would expected the mismatch to change the answer, but not give a huge degradation in the model solution (unless there is a quality / resolution issue between the 2 forcing products). When we have the ERA5 winds we should double check that it fixes things. 

I do think that the mismatch in forcing has a lot to do with it. GLORYS matches ERA5, so you'd expect that the surface temperatures / winds from the surface forcing at the boundary would be consistent with the lateral boundary conditions. In the rCM3 development we initially tried forcing the ocean with ACCESS-OM2 which is JRA forced, but with the atmospheric LBCs coming from ERA5 - this caused the boundary to completely blow up, and was resolved entirely by switching to GLORYS boundaries, matching the ERA5 atmospheric forcing for the UM

> In the meantime, I will have a closer look at the physics activated in the stencil as some of this may be different from the original stencil we were using in FMS (unless you have already ruled this out as the cause of the degradation?).

Yeah that could be contributing -  where did the physics in the stencil come from originally? I've had a quick look at the diff between the `MOM_parameter_doc.all` file for this run and an older one and there are some differences, but I suspect that a lot of them are due to model version changes (looks like many changed MOM defaults get flagged!). A quick check would be to do two runs on the current exe, and sub in the current MOM_input file from `demos/premade_run_dirs/common_files`, then have a look at the `MOM_parameter_doc.all` diff between the two tests

> 
> Are we able to merge into main before we
> 
I think it needs more testing first, and I'm hesitant to merge something to main when we know that the boundaries don't make sense! How long off is the ERA5 surface forcing support do you think? We could always just switch to the `setup_era5` method in the interim - this would just mean modifying the datm _in file to point to the locally regridded ERA5 files spat out by regional-mom6 which were used by the FMS runs

atteggiani:
> I think we may need to have this code in main before we can test it on the workflow.

Not necessarily. 
If you want to test the `regional-mom6` package in this PR, we can specify this branch when installing from git within in the dev environment specification `environment_dev.yml`:
```yaml
name: regional_mom6
channels:
  - accessnri
  - conda-forge
dependencies:
  - matplotlib
  - pip:
    - git+https://github.com/COSIMA/regional-mom6.git@rom3_setup
```

helenmacdonald:
Thanks @atteggiani! Can we please do that to start with so we can test?

ashjbarnes:
> > I think we may need to have this code in main before we can test it on the workflow.
> 
> Not necessarily. If you want to test the `regional-mom6` package in this PR, we can specify this branch when installing from git within in the dev environment specification `environment_dev.yml`:
> 
> ```yaml
> name: regional_mom6
> channels:
>   - accessnri
>   - conda-forge
> dependencies:
>   - matplotlib
>   - pip:
>     - git+https://github.com/COSIMA/regional-mom6.git@rom3_setup
> ```

Awesome yeah let's do that! What's the redeployment procedure? Would it be easy enough to trigger a rebuild of the env when we update the PR (not automatically via git actions, I know, but manually?)

atteggiani:
> **[ashjbarnes](/ashjbarnes) **

I left some instructions in the related [PR](https://github.com/ACCESS-NRI/containerised-environments-infra/pull/34)

Note for the moment only @helenmacdonald has permissions to do so.

helenmacdonald:
@ashjbarnes any chance we can please do another merge from main? The git fix (needed to progress the containerised env) went through just after your last one.

ashjbarnes:
> @ashjbarnes any chance we can please do another merge from main? The git fix (needed to progress the containerised env) went through just after your last one.

ah good catch! Done, hopefully that will work

manishvenu (approved):
On Tests Passing!

PR #267, opened 2026-01-24
Integration of mom6_forge & regional_mom6 (Remerged)

manishvenu:
Co-authored-by: @ashjbarnes 

Closes https://github.com/CROCODILE-CESM/CrocoDash/issues/183

**Summary**:

Integrate mom6_bathy into regional_mom6. There are NO front-end changes to this PR (the one exception is slight changes to what is printed out in the setup_bathymetry function). 

Shift the grid functions from regional_mom6 to mom6_bathy to *eventually* take advantage of the functions in mom6_bathy. For now, this PR just reproduces the user facing functions. 

This PR will add the mom6_bathy as a dependency (mom6_bathy is soon to be renamed and released on pypi). 

**Changes:**
expt.make_hgrid, make_vgrid, setup_bathymetry now use mom6_bathy grid, topo, and vgrid

**Advantages**:

There are a ton of advantages, but some highlights:

**Gain access to useful functions in mom6_bathy:**
1. Runoff Mapping
2. Chlorophyll File Generation for regional domains
3. Generate ESMF Mesh files (for NOUPC Coupling) and CICE grids with ease from a bathmetry file
5. Use the Topo Editor (an interactive widget for bathymetry editing)
6. EDIT: My favorite advantage is using the mom6_bathy fill method! [Here](https://i.imgur.com/NspviXV.png) is an example of how nice this fill method is when you use a low resolution product (like global CESM output).

**Regional_mom6/CROCODILE Advantages**:
7. Merge the grid generation advances CROCODILE & regional_mom6 make together 
8. mom6_bathy gains additional functions in generating hgrids and topos
9. More easily understand what a grid, topo, and vgrid are inside of regional_mom6-> they're objects instead of netcdf files
10. This will go a huge way to making these files more self-describing!

**Disadvantages/Considerations:**
1. Add another layer of complexity to regional_mom6
2. Currently, this development is on a branch of NCAR/mom6_bathy, which can make editing challenging. (I'd totally recommend COSIMA make a fork)
3. The big kicker is that this moves all of the grid generation out of regional_mom6 into mom6_bathy (Since mom6_bathy will be installed as a pip dependency, edits to grid generation would be PRs into the mom6_bathy repo)

Companion PR: https://github.com/NCAR/mom6_bathy/pull/34

**For reviewers/maintainers:**

1. Feature requests: This PR is ONLY to refactor where code is. Any major feature requests should go into issues! If you're reviewing this PR and notice something cool you want that's going to take a while to develop, make an issue. However, if it's small, I'm more than happy to do it.
2. **_BASELINES!:_** As briefly discussed a month or so ago, there's a definite need for baselines for both CrocoDash and regional_mom6. I'm thinking this (kind of major) PR is a great place to start doing that for the grids aspect. [Here is the way I'm doing them for now](https://github.com/CROCODILE-CESM/CrocoBaselines/tree/main/regional_mom6). If we're interested in starting them up on the COSIMA computer as well (which we definitely should!), please feel free to copy this code and try it out. I run it with:

```
python baseline_grid_generation.py <baseline_dir> --with-bathy # If you add the --with-bathy flag make sure you request additional resources
```
3. How to test this PR! You need to install mom6_bathy! Which can be done by [cloning it](https://github.com/NCAR/mom6_bathy) than `pip install -e .` in the cloned folder

4.  Instawins: Since a lot of these advantages will take some time & demo-writing, I dropped a demo for the TopoEditor, ESMF, & CICE files in this PR that should work, but will hold off on other demos for further PRs.

### DIFF
--- a/tests/test_regridding.py
+++ b/tests/test_regridding.py
@@ -9,6 +9,7 @@ from regional_mom6.regional_mom6 import segment
 import shutil
 from mom6_forge.grid import *
 
+
 # Not testing get_arakawa_c_points, coords, & create_regridder
 def test_smoke_untested_funcs(get_curvilinear_hgrid, generate_silly_vt_dataset):
     hgrid = get_curvilinear_hgrid
@@ -289,9 +290,9 @@ def test_regrid_velocity_tracers(toy_glorys_ds, tmp_path):
         ystart=2,
         leny=2,
         name="test",
-        type = "rectilinear_cartesian"
+        type="rectilinear_cartesian",
     )
-    hgrid = grid._supergrid.to_ds(name = grid.name, author="pytest")
+    hgrid = grid._supergrid.to_ds(name=grid.name, author="pytest")
     seg_name = "segment_001"
     outfolder = tmp_path / "inputdir"
     outfolder.mkdir()


ashjbarnes:
@manishvenu @angus-g , Do you think this is caused by the way that mom6_bathy is being installed directly via git? If we don't necessarily have to wait for mom6_bathy to be published on pypi then that would be ideal

angus-g:
> Do you think this is caused by the way that mom6_bathy is being installed directly via git? If we don't necessarily have to wait for mom6_bathy to be published on pypi then that would be ideal

Is "this" the failing conda install? Is there a log for that?

manishvenu:
> Thanks @manishvenu nah that's fine! I was just wondering if there was more I could do here or just let this sit until
> 
> 🔥 _mom6forge_ 🔥
> 
> was all published and ready to go. No rush at all this PR can wait :)

Ok sounds good. It also looks like the relaxing the requirements doesn't quite work, I'll track down this problem. 

Also mom6_bathy has code that looks like it won't work in 3.9 so maybe will have to change that!

manishvenu:
@ashjbarnes Looks like all the tests are passing in the CI with some minor changes to mom6_bathy if this PR needs to be merged. It's still very tentative, so still probably best to wait until we get it on pypi (or maybe being explicit this is a development version)!

This should mean the installation problems are fixed hopefully!

ashjbarnes:
One holdup is I need to make sure that we can deploy this properly on gadi with the dependency being a github repo. Hopefully it's fine, but there's a bit more faff than usual because can't just test it _on_ gadi, I also need to make sure that when the curated python environments (to which regional-mom6 belongs) are updated to new regional-mom6 version, it successfully picks up and installs the github repo. I've reached out to the ACCESS-NRI people for help and clarification :)

ashjbarnes (approved):
This works great - tested with OM2 forcing. The resultant input files are identical to the current main branch numerically, with some small changes to encoding. e.g., bathymetry now has coords named x/y instead of lat/lon. This is consistent with naming of the initial conditions anyway, and doesn't affect anything except how users plot them in notebooks. 

I've now tested this workflow with FMS runs too, and with some tweaks to the `run_fre_tools` function, this is now backwards compatible.

Contrary to what I said this morning, I'm happy for this to be merged! Turns out that regional-mom6 has already been kicked off the curated analysis env since we're still requiring numpy <2 so no-one will even know if we can't load mom6_forge properly! I'll work on the bespoke regional-mom6 deployment env next week. #

Issue #234, opened 2025-08-20
Support for numpy 2.0


rbeucher:
Hi all,

I am currently reviewing and updating some packages in the conda/analysis3 environment for the xp65 project on NCI. As part of this work, I am revisiting the packages that have been preventing us from moving to NumPy 2.0.

One of the packages flagged is regional-mom6. I wanted to check whether there is any major reason for this dependency to still hold us back from upgrading.

NumPy is now at version 2.3, so it seems like a good time to evaluate whether we can resolve these blockers and move the environment forward.

Has anyone already looked into regional-mom6 compatibility, or are there known issues that we should be aware of before testing an upgrade?

Thanks,
Romain

helenmacdonald:
I did a bit of digging and found [this](https://github.com/pangeo-data/xESMF/pull/377)
Which sounds like a fix to an issue similar to what you describe @angus-g?

angus-g:
Good find @helenmacdonald, that would probably fix it!

PR #268, opened 2026-01-26
updating numpy dependencies

helenmacdonald:
Testing to see if we are able to upgrade to a more recent version of numpy.
As requested [here](https://github.com/COSIMA/regional-mom6/issues/234) by @rbeucher


helenmacdonald:
Thanks @manishvenu - good to hear I am not the only one who encountered these issues!
>  I like your solution to force the output of the functions to be in a specific form (like numpy.float), or the reverse.

The problem is, that I did that and it didn't work 😞.   It was sucsessfull in forcing them to be the same type on my gadi environment, but was not for the testing here. I'm going to keep digging tomorrow. 

>I'd also recommend removing the numpy requirements from the environment-ci and pyproject file. If the environment solves numpy naturally above v2, that would be great (We also need to check if the ESMF regridding is still working).

Agreed, thanks. I will leave it forced to v2 until I have passed the testing, and then remove when I make this PR non draft. Do we need to write a test for the ESMF regridding - or will this be tested when I get past the current failing tests?

manishvenu:
@helenmacdonald @ashjbarnes I think the numpy is solving naturally and its passing all the tests now!
If you guys are good with it we should merge! I get a naturally solving environment of numpy ~2.4 and xesmf ~0.9

ashjbarnes (approved):

Issue #259, opened 2025-12-15
Validate OBC/IC Files Feature Request


manishvenu:
It would be great to have a function that validated the OBC/IC files to make sure they are formatted correctly (correct var names, FillValue, etc...)

PR #271, opened 2026-02-09
Add a Validation module

manishvenu:
Details:
 The very minimal start to a validation module -> hopefully with more time we can add more to it.

Fixes https://github.com/COSIMA/regional-mom6/issues/259

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -24,7 +24,7 @@ jobs:
         shell: bash -el {0}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v3


regional_mom6/regional_mom6.py
```diff
@@ -2799,7 +2821,7 @@ def regrid_velocity_tracers(
 
         # Here, keep in mind that 'var' keeps track of the mom6 variable names we want, and self.tracers[var]
         # will return the name of the variable from the original data
-
+        output_var_list = []
```
ashjbarnes: This PR is looking great! Just checking about some of the lines like here and similarly down below, where the main code is altered but doesn't immediately look like it's related to the validation stuff. Is this an additional change, or still to do with validation?

ashjbarnes (approved):
This looks good to me! Yeah I now understand what the changes to main are doing, should have been obvious oops

helenmacdonald:
I had a brief look over it and am happy for you to merge

PR #272, opened 2026-02-16
Remove Python versions 3.9 and 3.10 from testing

### LINKED ISSUE #270: Python 3.9 is past end-of-life!
Can we remove python 3.9 from the CI? @ashjbarnes @helenmacdonald 
Python version support: https://devguide.python.org/versions/

ashjbarnes:
Since 3.9 is at end of life and 3.10 is beyond support window, I've removed them from testing

closes #270


ashjbarnes:
oh good catch thanks!

PR #292, opened 2026-05-01
Protect regional_mom6 from breaking changes in mom6_forge versions

manishvenu:
mom6_forge uses classic Python versioning, so guarding against the minor version change protects regional-mom6 from breaking changes.


ashjbarnes (approved):
yes good idea!

PR #299, opened 2026-05-15
Turn Git Off

### LINKED ISSUE #289: read hgrid from file doesn't creat a self.grid object
The self.grid object nowadays points to a mom6_forge Grid object. This is populated when regional-mom6 creates a new grid via mom6_forge, but not when using the `hgrid_type = "from_file"` option. Just need to add a line here that sets `self.grid` to a Grid instance representing the hgrid.nc file being read in.

### LINKED ISSUE #295: Incorrect passphrase supplied
I am getting an error when running the latest version.

I am running this command:
```
expt.setup_bathymetry(
    bathymetry_path='path_to_bathy'', 
    longitude_coordinate_name='lon',
    latitude_coordinate_name='lat',
    vertical_coordinate_name='elevation',
    )
```
And get this error:
```
GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git commit -m COMMAND-Min Depth Edit
  stderr: 'error: Enter passphrase: Load key "path_to_key")": incorrect passphrase supplied to decrypt private key?
```

I believe that the line creating issues is:
```
self.repo.git.commit(
          "-m",
          f"{cmd_type.value}-{command_message if command_message is not None else ''}",
      )
```
See [here](https://github.com/NCAR/mom6_forge/blob/1f5a0d89d8e6a02cdd7823274325d32e5d65e16c/mom6_forge/command_manager.py#L63-L66) for context to the line.

I believe that the issue is that I have git signing activated and I need to input a passphrase whenever I do a commit. I need to keep the signed commits on due to access-nri policy but probably won't need to do signed commits for the commits being done here. I have done some digging, and I think we can turn off the signed commits for the commits done by mom_forge by passing in: `--no-gpg-sign`

For example, `payu` does it [here](https://github.com/payu-org/payu/blob/1a083723d112bb6e748a08f510f84a9f0a072a89/payu/runlog.py#L101):
```            
cmd = f'git commit --no-gpg-sign -am "{commit_msg}"'
```
@manishvenu  - any chance we can do this in mom6_forge?

manishvenu:
Fix #289 & Fix #295

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -321,6 +321,7 @@ class experiment:
                 hgrid_path = Path(hgrid_path)
             try:
                 self.hgrid = xr.open_dataset(hgrid_path)
+                self.grid = Grid.from_supergrid(hgrid_path)
                 self.longitude_extent = (
                     float(self.hgrid.x.min()),
                     float(self.hgrid.x.max()),
@@ -1433,6 +1434,7 @@ class experiment:
         self.topo = Topo(
             grid=self.grid,
             min_depth=self.minimum_depth,
+            git = False
         )
 
         self.topo.set_from_dataset(
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -1431,11 +1431,7 @@ class experiment:
         if regridding_method is None:
             regridding_method = self.regridding_method
 
-        self.topo = Topo(
-            grid=self.grid,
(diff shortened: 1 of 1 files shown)


ashjbarnes (approved):
looks good thanks @manishvenu !

manishvenu:
Thanks Ashley!

helenmacdonald (approved):
Thanks @manishvenu - I have checked and the issue is fixed!
I think you are good to merge

PR #300, opened 2026-05-27
Fix Longitude Slicer Import

manishvenu:


ashjbarnes (approved):
yes good catch I've run into this problem too! Wonder why it wasn't caught by tests? I just ran them again and they didn't get caught on import. I'll merge this rightaway since import isn't working without this fix

ashjbarnes:
Hold on, @manishvenu it looks like longitude_slicer isn't in utils but actually in [_source_bathy](https://github.com/NCAR/mom6_forge/blame/001a64f5e71eebb421063e713f135908e95c9cce/mom6_forge/_source_bathy.py#L244)

Is this a version thing? Like do we intend to have it in utils at some point and that's why it was importing from utils?

manishvenu:
Ashley should have mentioned - regional-mom6 uses the mom6-forge v0.0.x branch until we want to move to main which is currently working towards v0.1 (See my issue about the new bathy workflow, which has breaking changes). That branch still has longitude slicer in utils.

I'm not sure why tests don't catch it but it might be because we don't often straight import regional-mom6 (which might run init.py), we import specific files.

ashjbarnes:
Ah yeah this was my bad!! I was working on a PR into mom6_Forge but started from the main branch rather than from 0.0.x. Works fine now I'm on the right branch

PR #301, opened 2026-05-29
Chill overzealous info logging to debug, remove setup_logger

### LINKED ISSUE #262: Clean up logging output in regridding functions
I may have been * overzealous * in logging output from regridding functions. It's basically junk b/c it's too much. I lot of the "logging.info" statements should really be "logging.debug".

manishvenu:
Closes #262

- All `logging.info` calls in `regridding.py` (and one in `utils.py`) that describe internal processing steps are downgraded to `logging.debug`
- `setup_logger` in `utils.py` was a thin wrapper around `logging.getLogger` with dead handler/level logic; replaced with `logging.getLogger(__name__)` at each call site
- Recommendation from Claude: Added `NullHandler` to the package-level logger in `__init__.py` so regional-mom6 behaves as a library and does not emit output unless the caller configures logging (so things the user should see should be print statements)

### DIFF
--- a/regional_mom6/__init__.py
+++ b/regional_mom6/__init__.py
@@ -1,3 +1,7 @@
+import logging
+
+logging.getLogger(__name__).addHandler(logging.NullHandler())
+
 try:
     from ._version import __version__
 except ImportError:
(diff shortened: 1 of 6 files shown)


ashjbarnes (approved):
Excellent branch name

PR #302, opened 2026-05-29
Don't reuse regridding weights by default; expose regridders on segment

### LINKED ISSUE #296: Attempt a 100x100 domain centred on Lismore
Diagnosing an issue with setting up the four boundary conditions

### LINKED ISSUE #297: Regridding weights are saved
As pointed out by @manishvenu [here](https://github.com/COSIMA/regional-mom6/discussions/277), the regridding weights are saved

I think this might be the default behaviour of xesmf, but if we want we could delete the weights after regridding

The reason it's an issue is that if someone regenerates their grids then runs the workflow again, the weights will be reused and it will break with a pretty opaque error message as happened to @Rush74

Alternatively, we could continue to rely on the warning that Manish has already put into the regridding step. (see error in [this notebook](https://github.com/21centuryweather/UM_configuration_tools/blob/12-mom6-regridder_v2/notebooks/ACCESS-rOM3-demo.ipynb). That's how I diagnosed the problem so quickly. Perhaps once #262 is sorted out the fewer error messages will mean that the 'reuse weights' warning stands out more!

manishvenu:
Closes #297
Closes #296

Changes:

- `create_regridder` gains an explicit `reuse_weights=False` parameter. Previously weights were silently reused whenever the weights file already existed on disk, causing wrong results if the grid was regenerated. Now weights are always recomputed unless the caller explicitly opts in with `reuse_weights=True`.
- `segment.regrid_velocity_tracers` and `segment.regrid_tides` both accept an optional `regridders=` kwarg. When omitted, regridders are created as before and cached on `self.regridders` / `self.tidal_regridders`. Power users (e.g. CrocoDash running multiple time windows on the same grid) can pass those back in on subsequent calls to skip recomputation.
- Tidal u/v regridders now write weights files to `weights/` (previously nothing was written to disk for these).

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -2075,6 +2075,8 @@ class segment:
             self.bathymetry = None
         self.segment_name = segment_name
         self.repeat_year_forcing = repeat_year_forcing
+        self.regridders = None
+        self.tidal_regridders = None
 
     def regrid_velocity_tracers(
         self,
@@ -2086,6 +2088,7 @@ class segment:
         time_units="days",
         calendar="gregorian",
         fill_method=rgd.fill_missing_data,
+        regridders=None,
     ):
         """
         Cut out and interpolate the velocities and tracers.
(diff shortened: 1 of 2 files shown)


ashjbarnes (approved):
This is a really nice fix! So we can keep regridding generation on by default but no need to remake the weights if you know what you're doing

Issue #246, opened 2025-10-20
Initial Condition open_mf replace with open_dataset


manishvenu:
Because the IC fails on sufficently large domains (something to do with dask and interpolating_nans), for the first CrocoDash release, I went with a branch of regional_mom6: https://github.com/CROCODILE-CESM/regional-mom6/tree/making_ic_work_without_bugs that uses `open_dataset `instead for now. 

I think maybe that should be default behavior, just so that we don't run into those gotcha's -> I understand it may be less efficient, but I don't think anyone is running into resource limits with `open_dataset` (especially because it's only one day, so even globally is probably only 9 GB maximum?).

Issue #215, opened 2025-06-04
Possible bug in `setup_initial_condition` for large domain


mmr0:
I've been unable to generate the initial condition file using regional mom6 v1.0.1 for a large domain. The small tassie example works fine, but when I move to a bigger domain (~1000x1250x100) I get the following error:

`ValueError: dimension lon on 0th function argument to apply_ufunc with dask='parallelized' consists of multiple chunks, but is also a core dimension. To fix, either rechunk into a single array chunk along this dimension, i.e., ``.chunk(dict(lon=-1))``, or pass ``allow_rechunk=True`` in ``dask_gufunc_kwargs`` but beware that this may significantly increase memory usage.`

The notebook that I used is here https://github.com/mmr0/regional-recipe/blob/main/regional-mom6-forced-by-access-om2.ipynb
Using the environment analysis3-25.05

ping @navidcy

navidcy:
@angus-g, @ashjbarnes, @manishvenu any idea if we are chunking automatically here?

Or should we actually be passing the `allow_rechunk=True` here perhaps automatically here?

Or perhaps allow the user to pass `allow_rechunk` kwarg in `expt.setup_initial_condition` method and then that is passed down to xarray later when needed?

angus-g:
That's kind of a weird error message, but you can't actually (as far as I can tell) pass `allow_rechunk` to `interpolate_na`. The fix would probably have to be to open the IC dataset differently, it's just using a bare `open_mfdataset` call so chunking is automatic.

ashjbarnes:
Ahh I see - your workers only have 4.5gb of memory each, so dask HAS to chunk your files. By setting threads_per_worker=1 (which you should do), your memory is split equally between each worker. You should try set `n_workers = X` where X is a number of workers that means your workers have at least as much memory as your initial condition file. In this case, more than 8gb.

mmr0:
Hey! thanks all for looking at this. 
ic_unprocessed.nc is actually only 200MB or do (the grid size I cited was my model grid - I guess the unprocessed data is still at 1/10th degree). 
When I load with `xr.open_mfdataset` I can see that it is indeed chunked, despite being v. small
<img width="791" alt="Image" src="https://github.com/user-attachments/assets/dd6cd38b-cce1-4afb-8beb-d2574ff983bd" />

ashjbarnes:
Oh right, I guess the 8gb in the notebook refers to the raw rather than subset model data, my bad

[...]

PR #306, opened 2026-06-10
`setup_initial_condition` chunking fix

AidanJanney:
Depending on how initial condition data is prepared and handled, we can produce this error when using `setup_initial_condition`:

```
raise ValueError(
ValueError: dimension longitude on 0th function argument to apply_ufunc with dask='parallelized' consists of multiple chunks, but is also a core dimension. To fix, either rechunk into a single array chunk along this dimension, i.e., ``.chunk(dict(longitude=-1))``, or pass ``allow_rechunk=True`` in ``dask_gufunc_kwargs`` but beware that this may significantly increase memory usage.
```

See: https://github.com/COSIMA/regional-mom6/issues/215 and https://github.com/COSIMA/regional-mom6/issues/246.

We can address this by changing xr.open_mfdataset to xr.open_dataset. We expect to be opening single files for the initial conditions, and the largest initial conditions should still be less than 10GB so chunking is unnecessary. 

Other possible fixes: 
- Rechunk ic dataset after opening: `ic_raw = ic_raw.chunk({'longitude':-1, 'latitude':-1, 'depth':-1})`
- Use blank chunks with mfdataset: `xr.open_mfdataset(path, chunks={})`
These approaches have a chance of increasing computation by rechunking at computation time, but they do allow for handling of glob file paths or lists of file paths. However, this should not be an issue with the initial condition data. 

**Other changes:**
Minor changes to documentation and wrap strings in Path(str) when checking if a file exists.


regional_mom6/regional_mom6.py
```diff
@@ -654,7 +653,12 @@ def setup_initial_condition(
         reprocessed_var_map = apply_arakawa_grid_mapping(
             var_mapping=varnames, arakawa_grid=arakawa_grid
         )
-        ic_raw = xr.open_mfdataset(raw_ic_path)
+
+        if not Path(raw_ic_path).exists():
+            raise FileNotFoundError(
+                f"Boundary file not found at {raw_ic_path}. Please ensure that the files are named in the format `ic_unprocessed.nc`."
```
ashjbarnes:
```suggestion
                f"Initial condition file not found at {raw_ic_path}. Please ensure that the files are named in the format `ic_unprocessed.nc`."
```

ashjbarnes:
Awesome thanks @AidanJanney ! Yes I agree - seems more of a headache than it's worth to try and guess how to chunk things. Side note / unrelated to this PR, but are we still happy with the [imposed chunking on the BC](generate_layer_thickness)? I guess that the BCs _can_ be extremely large, so chunking temporally does make sense in this case if it's able to be done stably

ashjbarnes (approved):
Approving subject to typo fix

manishvenu:
> imposed chunking on the BC

Hey Ashley, the link you sent isn't working, but yeah -> I think the BC could likely get really large so we still want chunking for that. 

I think long-term we just need to change the way initial_condition is using that chunking. It really should work.

PR #308, opened 2026-07-15
m6f v0.3.0 Update

manishvenu:
Changes:

1. Update m6f version
2. Change constructor to take m6f objects
3. Wrap the m6f modules in there regional-mom6 counterparts, so the user doesn't see mom6-forge.
4. Call all the m6f objects "m6f_hgrid" "m6f_bathymetry" etc... to clarify them from the easily plottable xarray Dataset versions, which are properties that wrap those objects. So when you call expt.hgrid, it calls the "to_ds" equivalents on all the m6f objects.


ashjbarnes:
oh damn my bad what a dodgy merge

ashjbarnes (approved):

PR #315, opened 2026-07-17
PR: Make boundary/OBC forcing machinery standalone (Segment Class Expansion)

manishvenu:
**Summary**

Extracts the segment class from `regional_mom6.py` into a more expanded class (`segment.py`) that can be used independently of Experiment. Boundary now supports constructing arbitrary straight, index-aligned OBC segments (any axis, any index, any index range) — not just the 4 cardinal full edges — matching MOM6's actual OBC segment conventions.

**Key changes**

- New Segment class holds boundary geometry (lon/lat/angle/mask) and the regridding/rotation/masking/writing behavior as methods (regrid_velocity_tracers, regrid_tides, encode_tidal_files_and_output). Construct via Boundary.from_hgrid(hgrid, axis=, index=, index_range=, topo=) for any custom segment, or Boundary.cardinal(hgrid, orientation, ...) for the north/south/east/west cases. **What this means is we took out the `regridding.py coords` function, which was doing a lot of heavy lifting of boundary properties that belong in this class**
- **BIG CHANGE:** Reads hgrid["angle_dx"] directly instead of recomputing rotation angle (mom6_forge already computes this at grid-creation time); warns and falls back to zero angle if missing. This angle calculation responsibility is back with the user.
- Masking now uses mom6_forge.Topo.supergridmask directly as ground truth, with no regional-mom6 post-processing 
- Experiment internals updated to lazily build one Boundary per orientation (reusing self.segments), shared between ocean-state and tidal boundary setup instead of re-deriving it twice. No change to Experiment's public API.

- Added `demos/segment-basics-demo.ipynb`, a short standalone tour of building OBC segments via `Segment` alone (cardinal edges, interior/arbitrary lines, `from_lonlat`, `detect_open_cardinal_boundaries`).


manishvenu:
@ashjbarnes I think the easiest way to review this PR is to start by trying the demo notebook.

ashjbarnes (approved):
should make a new tag when this is merged like +0.0.1

manishvenu:
@ashjbarnes Thanks! Want me to make the tag?

ashjbarnes:
Thanks all good I'll do it!

PR #325, opened 2026-08-27
Add an opt-in ignore_degenerate option to regridder construction

manishvenu:
## Problem

`xe.Regridder()` aborts the entire regrid when ESMF meets a degenerate (duplicate or collapsed) source cell:

```
ESMC_FieldRegridStore failed with rc=506    (Degenerate Element Detected)
```

Some global source datasets have such cells near the poles — GLORYS, for instance, at its southernmost latitudes — so regridding OBC segments for a polar domain fails outright, even though only a handful of unusable source cells are involved. There's currently no way to get past it from `regional-mom6`'s API.

## Change

ESMF can skip those cells rather than fail (`ignore_degenerate`), so this exposes it as an **opt-in** parameter, threaded down the path that builds the segment regridders:

- `regridding.create_regridder(..., ignore_degenerate=False)` → passed to `xe.Regridder`
- `create_vt_regridders(..., ignore_degenerate=False)` → passed to all three `create_regridder` calls
- `segment.regrid_velocity_tracers(..., ignore_degenerate=False)` → passed to `create_vt_regridders`

Purely additive: 25 inserted lines, nothing deleted, no existing call site changed.

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -2285,6 +2285,7 @@ class segment:
         calendar="gregorian",
         fill_method=rgd.fill_missing_data,
         regridders=None,
+        ignore_degenerate=False,
     ):
         """
         Cut out and interpolate the velocities and tracers.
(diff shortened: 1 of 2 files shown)


ashjbarnes (approved):
Looks good to me, good find. Later we could set it to default to True if we find no issues, but I like that it retains current behaviour by default for now

PR #327, opened 2026-09-04
replace data fill function with simpler, more robust version

### LINKED ISSUE #326: The "fill_missing_data" function doesn't fill vertically, only fills 1 direction horizontally
Currently, the [fill_missing_data](https://github.com/COSIMA/regional-mom6/blob/fe082e0349c856cae36a8f45f4b66a3397e89da7/regional_mom6/regridding.py#L281) function has two problems:

1. If passed a list of multiple z dims (as is the case for all boundaries) it will fill the nans with 0 on the first loop, making the rest redundant
2. It only fills horizontally in one direction. This means that if there's coastline to one side but no the other, some coast won't get filled. This will later be filled with 0s due to the first issue by accident, but not actually filled with meaningful values

Additionally there's an issue with how the `fill_missing_data` function is [called](https://github.com/COSIMA/regional-mom6/blob/fe082e0349c856cae36a8f45f4b66a3397e89da7/regional_mom6/regional_mom6.py#L2514) in the main regional_mom6 code. It _only_ passes a single z dimension anyway (in this case, the z dim of the original dataset, depth in the case of GLORYS). This means that even if 1 and 2 were fixed, the actual data never gets filled anyway. The segment's vertical dims are called `nz_segment_00X_variablename`, not `depth`. 

The upshot is that salinity values get filled with 0 rather than sensible values. When this happens at the seafloor it causes issues!

ashjbarnes:
Closes #326 

This suggested fix massively simplifies the fill missing data method. Looking at the way it's actually used in the code there are only two use cases:

1. Only fill along a single dimension (working)
2. Fill across x dimension and then along every dimension (broken)

This fix retains 1 exactly (important for the tidal regridding). 

Case 2 is only used once, and that's in the velocity/tracer regridding. The goal here is to fill in all of the nans; it doesn't really matter what order that's done in. Anything is an improvement to the current "fill with 0" bug anyway!

### DIFF
--- a/regional_mom6/regional_mom6.py
+++ b/regional_mom6/regional_mom6.py
@@ -2510,8 +2510,7 @@ class segment:
         # fill in NaNs
         segment_out = fill_method(
             segment_out,
-            xdim=f"{coords.attrs['parallel']}_{self.segment_name}",
-            zdim=reprocessed_var_map["depth_coord"],
+            dim="all",
         )
 
         # Overwrite the actual lat/lon values in the dimensions, replace with incrementing integers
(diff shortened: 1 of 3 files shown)


manishvenu (approved):
Cheers! looks good to me, this change means that any fill_method that is *passed in* is going to need to be wrapped to either have dim="all" or one dimension, which is not a problem, just a different way of doing things.

PR #330, opened 2026-09-15
complete renaming of vcoord to vgrid

ashjbarnes:
This renaming brings rmom6's terminology inline with mom6_Forge who have a vgrid object. 

Only affects people trying to read in a vgrid/coord from file, since rmom6 currently writes as vcoord.nc but expexts to read as vgrid.nc


ashjbarnes:
omg I didn't mean to click on copilot go away

helenmacdonald:
I was wondering what was going on when I saw the copilot emails!

Note that the global access-om3 models actually use `ocean_vgrid.nc` - but it does make more sense for this package to align with mom6_forge so happy to keep `vgrid.nc`.

helenmacdonald (approved):

ashjbarnes:
Thanks @helenmacdonald ! Yes ocean vgrid at least is a bit more compatible than vcoord. I think I named it that way originally to remind people that it's a _target_ grid/coordinate rather than fixed like the hgrid but I think it's very confusing to use different terminology

